PR #38, opened 2023-08-08
Skip locks for schema-only backups via config settings

MedvedewEM:
Now in config we can skip locks while schema-only backups.
Setting path in config - `backup: skip_lock_for_schema_only: backup`

### DIFF
--- a/ch_backup/ch_backup.py
+++ b/ch_backup/ch_backup.py
@@ -162,8 +162,10 @@ class ClickhouseBackup:
             ", ".join(map(lambda db: db.name, databases)),
         )
 
+        skip_lock = self._check_schema_only_backup_skip_lock(sources)
+
         try:
-            with self._context.locker():
+            with self._context.locker(disabled=skip_lock):
                 if sources.access:
                     self._access_backup_manager.backup(self._context)
                 if sources.udf:
@@ -581,3 +583,13 @@ class ClickhouseBackup:
             return True
 
         return False
+
+    def _check_schema_only_backup_skip_lock(self, sources: BackupSources) -> bool:
+        if not sources.schema_only:
+            return False
+
+        skip_lock = self._context.config.get("skip_lock_for_schema_only", None)
+        if not skip_lock:
+            return False
+
+        return skip_lock.get("backup", False)
--- a/ch_backup/logic/lock_manager.py
+++ b/ch_backup/logic/lock_manager.py
@@ -9,6 +9,7 @@ from typing import IO, Optional, Type
 
 from kazoo.recipe.lock import Lock
 
+from ch_backup import logging
 from ch_backup.zookeeper.zookeeper import ZookeeperClient, ZookeeperCTL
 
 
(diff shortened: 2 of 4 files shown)


ch_backup/ch_backup.py
```diff
@@ -581,3 +584,13 @@ def _check_min_interval(self, last_backup: BackupMetadata, force: bool) -> bool:
...
+
+    def _check_schema_only_backup_skip_lock(self, sources: BackupSources) -> bool:
+        if not sources.schema_only:
+            return False
+
+        skip_lock = self._context.config.get("skip_lock_for_schema_only", None)
+        if not skip_lock:
+            return False
+
+        return skip_lock.get("backup", False)
```
ianton-ru, line 596: Why just `skip_lock_for_schema_only: True` is not enough, what is purpose of subnode `backup`?

ch_backup/ch_backup.py
```diff
@@ -156,14 +156,17 @@ def backup(
 
         self._context.backup_layout.upload_backup_metadata(self._context.backup_meta)
 
+        skip_lock = self._check_schema_only_backup_skip_lock(sources)
+        logging.debug("Lock will be skipped for backup: {skip_lock}")
```
ianton-ru, line 160: I suggest to move line "Lock will be skipped" after "Starting backup". May be better not here but inside `locker` where we actually skip lock.

ianton-ru (changes requested):

ch_backup/ch_backup.py
```diff
@@ -156,14 +156,17 @@ def backup(
...
+
         logging.debug(
             'Starting backup "%s" for databases: %s',
             self._context.backup_meta.name,
             ", ".join(map(lambda db: db.name, databases)),
         )
 
         try:
-            with self._context.locker():
+            with self._context.locker(skip_lock=skip_lock):
```
aalexfvk, line 169: I would propose not to add parameter for locker. Why does one need a locker that does not lock anything ).
But do something like this:
```
if skip_lock:
    locker = contextlib.nullcontext()
    logging.debug("Lock will be skipped for backup")
else:
    locker = self._context.locker()

with locker:
        # Do something
```

ch_backup/ch_backup.py
MedvedewEM: Lock is also taken in methods:
backup, restore, delete, purge, restore_access_control, fix_admin_user

Maybe later we will need to add the separate settings for this methods too.

ch_backup/ch_backup.py
MedvedewEM: What is about changing parameter's name: `with self._context.locker(disabled=skip_lock):` ?

For me it seems more concisely - disabled locker makes some sense now.
And in case of necessity there will not be such if/else construction in other places.

aalexfvk: Okay, I don't mind. My option isn't dramatically better. But anyway log message should be conditional.

aalexfvk (approved):

ianton-ru (approved):

PR #46, opened 2023-08-15
Enable version command when clickhouse server is unavaliable.

MikhailBurdukov:
Enabled using `version` without ClickHouse server. Made BackupContext loading in lazily way.


ch_backup/ch_backup.py
```diff
@@ -31,10 +31,39 @@
...
+class ContextLazyLoader:
+    """
+    Lazy loader for BackupContext
+    """
+
+    def __init__(self, config: Config) -> None:
+        self._config = config
+
+    @cached_property
+    def context(self) -> BackupContext:
```
aalexfvk, line 47: Why not just move this method in `ClickhouseBackup` class. And in `reload_config()` just clear its cache
`self.context.fget.cache_clear()`.

`ContextLazyLoader` class seems too specialized.

images/clickhousenozk/config/supervisor/conf.d/clickhouse.conf
```diff
@@ -2,7 +2,7 @@
 command=/usr/bin/clickhouse-server --config /etc/clickhouse-server/config.xml
 process_name=%(program_name)s
 autostart=true
-autorestart=true
+autorestart=false
```
aalexfvk, line 5: Maybe "unexpected" with exitcodes (0 by default) is more suitable ? To cover failing terminations.

tests/integration/features/commands_without_chserver.feature
```diff
@@ -0,0 +1,12 @@
+
```
aalexfvk, line 1: Suggested naming like `backup_other_commands` or `backup_misc_commands`

tests/integration/steps/ch_backup.py
```diff
@@ -189,3 +189,12 @@ def step_check_s3_backup_directory(context, node):
         equal_to(expected_directories),
         f"Actual backup directories = {actual_directories}, expected {expected_directories}",
     )
+
+
+@then("we got a vaild ch-backup version on {node:w}")
```
aalexfvk, line 194: valid

tests/integration/modules/ch_backup.py
```diff
@@ -198,6 +198,9 @@ def __init__(self, context: ContextT, node_name: str, timeout: int = 300) -> Non
         protocol = context.ch_backup["protocol"]
         self._cmd_base = f"timeout {timeout} {CH_BACKUP_CLI_PATH} --protocol {protocol} --insecure  --config {self._config_path}"
 
+    def get_version(self) -> str:
```
ianton-ru, line 201: What a difference with `version` method below? Looks like the same.

ch_backup/util.py
```diff
@@ -369,3 +369,12 @@ def dataclass_from_dict(type_: Type[T], data: dict) -> T:
     """
     class_fields = {f.name for f in data_fields(type_)}
     return type_(**{k: v for k, v in data.items() if k in class_fields})  # type: ignore[call-arg]
+
+
+def cached_property(f):
```
ianton-ru, line 374: How long will this cache properties? What if some properties change, how to invalidate cache?

ianton-ru (changes requested):

ch_backup/util.py
aalexfvk: I no longer like this approach, although I had previously advised it. It does not fit this case very well.

It caches forever and can lead to memory leaks when creating a large number of objects. It will keep these objects alive.

I propose to ged rid of generalization and use simple ad hoc property with saving context in _variable. And setting it to None when clear is needed.

Or [this](https://github.com/bottlepy/bottle/blob/3cc41c83545d43d3577bf88cd160dd64ab146359/bottle.py#L175) approach which is more suitable for this case. For invalidating it is used `del self.my_cached_attribute`.

ch_backup/ch_backup.py
MikhailBurdukov: Fixed

tests/integration/features/commands_without_chserver.feature
MikhailBurdukov: Fixed

tests/integration/modules/ch_backup.py
MikhailBurdukov: Yes, you right. Missed it. Removed it in the last commit.

ch_backup/util.py
MikhailBurdukov: Fixed it. Made it as Alexey  proposed with cached_property and invalidating to avoid memory leaks.
>Or [this](https://github.com/bottlepy/bottle/blob/3cc41c83545d43d3577bf88cd160dd64ab146359/bottle.py#L175) approach which is more suitable for this case than current one. For invalidating it is used del self.my_cached_attribute.

tests/integration/features/commands_without_chserver.feature
aalexfvk: This doesn't seem to have changed.

MikhailBurdukov: Yea, sorry. Renamed in the last commit.

aalexfvk (approved):

MikhailBurdukov:
@ianton-ru

ianton-ru (approved):

PR #47, opened 2023-08-18
Set `force_drop_table` flag before deleting tables while restoring

aalexfvk:


tests/integration/features/backup_restore.feature
```diff
@@ -153,3 +153,32 @@ Feature: Backup & Restore
...
+
+    INSERT INTO test_db.test_table SELECT number % 2, number, number FROM system.numbers LIMIT 100;
+    """
+    And we put following info in "/etc/clickhouse-server/conf.d/max_table_size_to_drop.xml" at clickhouse02
+    """
+    <yandex>
+      <max_table_size_to_drop>1</max_table_size_to_drop>
+    </yandex>
+    """
+    And we restore clickhouse backup #0 to clickhouse02
```
Alex-Burmak, line 183: ClickHouse configs are reloaded asynchronously. So backup restore can happen before config reload.

Suggest to invoke `SYSTEM RELOAD CONFIG` after updating config files.

aalexfvk: Of course. Reloading has been added.

Alex-Burmak (approved):

PR #49, opened 2023-08-22
Latest backup in any state should not be deleted while partial delete command

MedvedewEM:
**Example:**
```      
backup100 CREATING
backup99 CREATED
backup98 FAILED
...
backup0 CREATED
```

`backup100` - newest
`backup0` - oldest

**Given:**
We want to delete oldest one, but backup100 is still creating.

**Actions:**
- we run `ch-backup delete --purge-partial backup0`
- we compute `deleting_backups` and `retained_backups` and collect deduplication data
- `backup100`'s metadata is loaded with `state=CREATING`, so we think it could be deleted/partially deleted as well
- while we downloading **other** backups metadata (it could take a while) `backup0` is changing state to `CREATED`
- delete command finishes metadata downloading
- `backup100` will be deleted, but it was created right now

**Fix**
Skip newest backup while purge-partial flag in delete command.

Btw, lock on backup and delete commands does not help there because metadata loading works outside of lock scope.
Also newest backup could be useful for future's backups because of deduplication.

### DIFF
--- a/ch_backup/ch_backup.py
+++ b/ch_backup/ch_backup.py
@@ -334,13 +334,15 @@ class ClickhouseBackup:
         found = False
         deleting_backups = []
         retained_backups = []
-        for backup in self._context.backup_layout.get_backups(use_light_meta=True):
+        for i, backup in enumerate(
+            self._context.backup_layout.get_backups(use_light_meta=True)
+        ):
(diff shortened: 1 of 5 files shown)


MedvedewEM:
https://github.com/yandex/ch-backup/pull/50

PR #50, opened 2023-08-22
Latest backup in any state should not be deleted while partial delete command

MedvedewEM:
**Example:**
```      
backup100 CREATING
backup99 CREATED
backup98 FAILED
...
backup0 CREATED
```

`backup100` - newest
`backup0` - oldest

**Given:**
We want to delete oldest one, but backup100 is still creating.

**Actions:**
- we run `ch-backup delete --purge-partial backup0`
- we compute `deleting_backups` and `retained_backups` and collect deduplication data
- `backup100`'s metadata is loaded with `state=CREATING`, so we think it could be deleted/partially deleted as well
- while we downloading **other** backups metadata (it could take a while) `backup0` is changing state to `CREATED`
- delete command finishes metadata downloading
- `backup100` will be deleted, but it was created right now

**Fix**
Skip newest backup while purge-partial flag in delete command.

Btw, lock on backup and delete commands does not help there because metadata loading works outside of lock scope.
Also newest backup could be useful for future's backups because of deduplication.


Alex-Burmak:
LGTM

> Btw, lock on backup and delete commands does not help there because metadata loading works outside of lock scope.

It's worth to revisit this logic. Probably we should reload metadata after acquiring lock or take lock before loading metadata.

But it doesn't supersede the changes from the current PR. As already mentioned in the PR description, it's worth to keep the newest backup for deduplication purposes regardless of its state.

Alex-Burmak (approved):

MedvedewEM:
I've investigated more precisely your suggestions:
1. Take lock before loading metadata - it is easy, I can do it right now, but it has one obvious disadvantage: it will lock command for a longer time.
2. Reload metadata after acquiring lock - we reload full metadata about backup [there](https://github.com/yandex/ch-backup/blob/main/ch_backup/ch_backup.py#L453), so we could move out this line to [here](https://github.com/yandex/ch-backup/blob/main/ch_backup/ch_backup.py#L364) (the same for purge command).
Then check backup status again and if it differs from previous one (and only if it is CREATED? because `CREATING -> FAILED` is ok to delete) - skip deleting this backup.
It is more complex logic and maybe needs more investigation, so I am not ready to do it in current PR.

What do you think about second option and to do it separately?
@Alex-Burmak

Alex-Burmak:
I think we can go with the first option. Even it's a trivial change I propose to do it in a separate PR.

GrigoryPervakov:
It's ok to hold lock for a longer time, before this feature we holder it for even longer period

aalexfvk (approved):

PR #59, opened 2023-08-31
Lock feature test fix.

MikhailBurdukov:
Two more tests ( [1](https://github.com/yandex/ch-backup/blob/00ff3e274657bfc8f23659d7bcff6e64b57edfdb/tests/integration/features/lock.feature#L11) and  [2](https://github.com/yandex/ch-backup/blob/00ff3e274657bfc8f23659d7bcff6e64b57edfdb/tests/integration/features/lock.feature#L26))  will be fixed and committed in this [PR](https://github.com/yandex/ch-backup/pull/57). Because they are broken rn.

### DIFF
--- a/tests/integration/features/lock.feature
+++ b/tests/integration/features/lock.feature
@@ -5,57 +5,24 @@ Feature: Lock
     And a working s3
     And a working zookeeper on zookeeper01
     And a working clickhouse on clickhouse01
-    And a working clickhouse on clickhouse02
 
-  @require_version_22.3
-  Scenario: No backup created while zk lock
-    When ch-backup configuration on clickhouse01
-    """
-    backup:
-        skip_lock_for_schema_only:
-            backup: false
-    lock:
-        zk_flock: true
-        zk_flock_path: /ch_backup/zk_flock_path
-    """
-    And on zookeeper01 we create /ch_backup/zk_flock_path
-    When we create clickhouse01 clickhouse backup
-    Then we got no backups on clickhouse01
-
-  @require_version_22.3
-  Scenario: No "schema-only" backup created while zk lock
-    When ch-backup configuration on clickhouse01
-    """
-    backup:
-        skip_lock_for_schema_only:
-            backup: false
-    lock:
-        zk_flock: true
-        zk_flock_path: /ch_backup/zk_flock_path
-    """
-    And on zookeeper01 we create /ch_backup/zk_flock_path
-    When we create clickhouse01 clickhouse backup
-    """
-    schema_only: true
-    """
-    Then we got no backups on clickhouse01
-
-  @require_version_22.3
-  Scenario: Skip zk lock while creating "schema-only" backup
-    When ch-backup configuration on clickhouse01
+  Scenario: Skip lock while creating "schema-only" backup
+    Given ch-backup configuration on clickhouse01
     """
     backup:
         skip_lock_for_schema_only:
             backup: true
     lock:
-        zk_flock: true
-        zk_flock_path: /ch_backup/zk_flock_path
+        zk_flock: false
+        flock: true
+        flock_path: /tmp/flock.lock
     """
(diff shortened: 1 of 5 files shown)


tests/integration/features/lock.feature
```diff
@@ -5,57 +5,24 @@ Feature: Lock
     And a working s3
     And a working zookeeper on zookeeper01
     And a working clickhouse on clickhouse01
-    And a working clickhouse on clickhouse02
 
-  @require_version_22.3
```
aalexfvk, line 10: Why was these tests deleted ?

tests/integration/features/lock.feature
```diff
@@ -5,57 +5,24 @@ Feature: Lock
...
+  Scenario: Skip lock while creating "schema-only" backup
+    Given ch-backup configuration on clickhouse01
     """
     backup:
         skip_lock_for_schema_only:
             backup: true
     lock:
-        zk_flock: true
-        zk_flock_path: /ch_backup/zk_flock_path
+        zk_flock: false
```
aalexfvk, line 16: Why did you changed zk lock to file lock in this test ? Maybe it makes sense just to add new test for testing via file lock

tests/integration/features/lock.feature
MikhailBurdukov: All these test are disabled rn (`lock.feature` doesn't included in `ch_backup.featureset`), also they are should not pass with current master branch.

In deleted tests we trying to create backup when [lock](https://github.com/yandex/ch-backup/blob/00ff3e274657bfc8f23659d7bcff6e64b57edfdb/ch_backup/ch_backup.py#L169) is already captured by other process and assuming that there are no backups. But there is a bug, when the lock acquired, we will upload only metadata and the backup will hang in the `creating` state. Finally we will have 1 backup instead of 0. 
This [PR](https://github.com/yandex/ch-backup/blob/00ff3e274657bfc8f23659d7bcff6e64b57edfdb/ch_backup/ch_backup.py#L161) resolving such problem.  
And deleted tests will be added in it, when this PR will be merged.

tests/integration/features/lock.feature
MikhailBurdukov: Because I haven't found the better way how to acquire the lock from outside the ch-backup.

The problem that I have  faced with `zk_flock`:
For locking we are using [`Kazoo`](https://kazoo.readthedocs.io/en/latest/_modules/kazoo/recipe/lock.html) lib and it requires to have alive  connection with zookeeper throughout the entire time while holding the lock. So if capture it  in the same way as in [`lock_manager`](https://github.com/yandex/ch-backup/blob/00ff3e274657bfc8f23659d7bcff6e64b57edfdb/ch_backup/logic/lock_manager.py#L98) inside the test and run ch-backup, connection with zookeeper can not be establish, because we already have alive connection on the same node with it.

BTW looks like for these scenarios does not matter in which way we acquire the lock, we tests how the `backup` cmd will works in such environment.

MikhailBurdukov: Also current [method](https://github.com/yandex/ch-backup/blob/00ff3e274657bfc8f23659d7bcff6e64b57edfdb/tests/integration/features/lock.feature#L21) for lock acquiring is incorrect and doesn't work at all.

aalexfvk: Ok, I see
Nevertheless, It would be better to use zk lock instead of file, because we use it in production.
I can propose a couple of approaches:
1. Not very good.
Use sentence like 
`When on zookeeper01 we create /ch_backup/zk_flock_path/__lock__-0000000000`
It works but exposes some internal details.
2. Better.
Write steps using `LockManager` and saving it in the behave `context` object for passing to next steps for releasing.
But `connection with zookeeper can not be establish, because we already have alive connection on the same node with it.` might prevent this. I'm not sure, because not tested.

aalexfvk (approved):

PR #65, opened 2023-09-06
Do not redownload parts that was downloaded, but not attached

GrigoryPervakov:
Change-Id: I13aba57ca58b75192481897673b6a705d7cfe0c8

### DIFF
--- a/ch_backup/logic/table.py
+++ b/ch_backup/logic/table.py
@@ -16,6 +16,7 @@ from ch_backup.backup.deduplication import (
     deduplicate_part,
 )
 from ch_backup.backup.metadata import PartMetadata, TableMetadata
+from ch_backup.backup.restore_context import PartState
 from ch_backup.backup_context import BackupContext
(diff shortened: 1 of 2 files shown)


ch_backup/logic/table.py
```diff
@@ -605,6 +612,7 @@ def _restore_data(
                                 context.backup_meta, part, fs_part_path
                             )
 
+                        context.restore_context.add_part(part, PartState.DOWNLOADED)
```
aalexfvk, line 615: Actually,  the part has been not downloaded yet. Because `download_data_part()` downloads parts asynchronously. All parts will be downloaded after `context.backup_layout.wait(keep_going)`.

GrigoryPervakov: Oh, forgot about it, I'm fixing redundant copying on restart for cloud storage

aalexfvk (approved):

PR #67, opened 2023-09-08
Fix Python version detection in Makefile

Alex-Burmak:


Makefile
```diff
@@ -163,7 +163,16 @@ ch_backup/version.txt:
 
 
 .PHONY: install-deps
-install-deps: .install-deps
+install-deps: check-environment .install-deps
+
+.PHONY: check-environment
+check-environment:
+	@if ! command -v ${PYTHON} > /dev/null; then \
```
aalexfvk, line 170: I'm not sure about `/dev/null` in makefile. For example, it is not mounted in chroot created in teamcity by pbuilder for building deb package and we will get an error.

Alex-Burmak: Rewrote the check to not use `/dev/null`.

aalexfvk (approved):

PR #68, opened 2023-09-11
Take a lock on wider scope in `delete` and `purge` cmd.

### LINKED ISSUE #51: Take a lock on collecting backups info for deduplication while deleting with purge-partial parameter
See discussion for details https://github.com/yandex/ch-backup/pull/50#issuecomment-1689592502

**What is expected to do**
1. Take a lock on wider scope in `ch_backup.delete(...)` method
2. Check that calling `_validate_and_resolve_name` at the begging of delete command does not cache backups and it does not have collision with changes in 1.
3. After 1. changes: should we reload backup none-light backup's info in `ch_backup._delete(...)` method? Could we load it immediately once?
4. I think it is the same for `purge` command

MikhailBurdukov:
- Moved lock acquisition at the begging of `delete`  and `purge` command.
- Made it so that the non-light metadata is loaded once for all backups .  

 Resolves #51.

### DIFF
--- a/ch_backup/backup/deduplication.py
+++ b/ch_backup/backup/deduplication.py
@@ -284,8 +284,9 @@ def collect_dedup_references_for_backup_deletion(
     """
     dedup_refences = collect_dedup_references_for_batch_backup_deletion(
         layout=layout,
-        retained_backups_with_light_meta=retained_backups_with_light_meta,
-        deleting_backups_with_light_meta=[deleting_backup_with_light_meta],
+        retained_backups=retained_backups_with_light_meta,
+        deleting_backups=[deleting_backup_with_light_meta],
+        backups_with_light_meta=True,
     )
 
     return dedup_refences[deleting_backup_with_light_meta.name]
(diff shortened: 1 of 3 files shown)


aalexfvk (approved):
It looks okay to me, but let someone else look at it.

ch_backup/backup/deduplication.py
```diff
@@ -284,29 +284,30 @@ def collect_dedup_references_for_backup_deletion(
     """
     dedup_refences = collect_dedup_references_for_batch_backup_deletion(
         layout=layout,
-        retained_backups_with_light_meta=retained_backups_with_light_meta,
-        deleting_backups_with_light_meta=[deleting_backup_with_light_meta],
+        retained_backups=retained_backups_with_light_meta,
```
MedvedewEM, line 287: Do we need method `collect_dedup_references_for_backup_deletion`?

Seems like it is only one method with `backups_with_light_meta=True`, except unittests - maybe we do not need this parameter at all?

MikhailBurdukov: > Do we need method collect_dedup_references_for_backup_deletion?
Seem like not, it is unused method. Will remove it. Well, therefore we can remove `backups_with_light_meta`.

MikhailBurdukov: Done

MedvedewEM (approved):

PR #72, opened 2023-09-25
Add external run action test-integration

Varagon007:
Added possibility to run test-integration using external trigger.

### DIFF
--- a/.github/workflows/ext_run_test_integration.yml
+++ b/.github/workflows/ext_run_test_integration.yml
@@ -1,0 +1,33 @@
+name: external_run_test-integration
+
+run-name: ${{ inputs.prefix }}_${{github.workflow}}_${{ inputs.uuid }}
+
+on:
+  workflow_dispatch:
+    inputs:
+      prefix:
+        description: 'Prefix runtime workflow'
+        required: false
+        type: string
+        default: ""
+      uuid:
+        description: 'UUID that can correct understand start'
+        required: false
+        type: string
+        default: ""
+      clickhouse_version:
+        description: 'Clickhouse version'
+        required: true
+        type: string
+
+jobs:
+  test_integration:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: set up python
+        uses: actions/setup-python@v4
+        with:
+            python-version: "3.11"
+      - name: run integration tests
+        run: CLICKHOUSE_VERSION=${{ inputs.clickhouse_version }} make test-integration


.github/workflows/ext_run_test_integration.yml
```diff
@@ -0,0 +1,33 @@
+name: external_run_test-integration
```
Alex-Burmak, line 1: Propose to name the workflow `test_clickhouse_version`.

.github/workflows/ext_run_test_integration.yml
```diff
@@ -0,0 +1,33 @@
...
+        default: ""
+      uuid:
+        description: 'UUID that can correct understand start'
+        required: false
+        type: string
+        default: ""
+      clickhouse_version:
+        description: 'Clickhouse version'
+        required: true
+        type: string
```
Alex-Burmak, line 21: Why do we need `prefix` and `uuid` parameters? Is it not enough to have only `clickhouse_version`?

.github/workflows/ext_run_test_integration.yml
Varagon007: Changed

aalexfvk (approved):

.github/workflows/ext_run_test_integration.yml
Alex-Burmak: It seems we can use `inputs.clickhouse_version` in `run-name` and get rid of `input.prefix`.

Also suggest to make id parameter more abstract - just `id` without any implications on format.

```
run-name: ${{ github.workflow }}_${{ inputs.clickhouse_version }}_${{ inputs.id || github.run_number }}

on:
  workflow_dispatch:
    inputs:
      clickhouse_version:
        description: 'ClickHouse version'
        required: true
        type: string
      id:
        description: 'Run identifier'
        required: false
        type: string
        default: ""
```
?

.github/workflows/ext_run_test_integration.yml
Alex-Burmak: Let's also change workflow file name for consistency.

.github/workflows/ext_run_test_integration.yml
Varagon007: Yes, it is possible. I will change Tasklet config in out side too. It is no problem. I will change.

Varagon007: I need time for think. Because this formula of "run-name" must me used in our side in Tasklets. Tasklets will be universal. May be I will add optional field in Tasklet for custom.

.github/workflows/ext_run_test_integration.yml
Varagon007: Done

.github/workflows/ext_run_test_integration.yml
Varagon007: Corrected.

PR #73, opened 2023-09-27
Add configurable timeout for zk_lock retry

mdyuldina:
1. Now we try to acquire zk_lock in blocking mode
2. Added configurable timeout after which acquire of lock will be unsuccessful

### DIFF
--- a/ch_backup/logic/lock_manager.py
+++ b/ch_backup/logic/lock_manager.py
@@ -118,7 +118,7 @@ class LockManager:
             raise RuntimeError("ZK flock enabled, but zookeeper is not configured")
 
 
-def fail_on_lock_acquiring(force_exit: bool, e: Optional[Exception] = None):
+def fail_on_lock_acquiring(force_exit: bool, e: Optional[Exception] = None) -> None:
     """
     Determines how to exit from the lock acquisition process
     """
--- a/ch_backup/logic/lock_manager.py
+++ b/ch_backup/logic/lock_manager.py
@@ -41,6 +41,7 @@ class LockManager:
         self._exitcode = lock_conf.get("exitcode")
         self._distributed = zk_ctl is None
         self._disabled = False
+        self._lock_timeout = lock_conf.get("lock_timeout")
 
     def __call__(self, distributed: bool = True, disabled: bool = False):  # type: ignore
         """
@@ -106,7 +107,7 @@ class LockManager:
             if not client.exists(self._process_zk_lockfile_path):
                 client.create(self._process_zk_lockfile_path, makepath=True)
             self._zk_lock = client.Lock(self._process_zk_lockfile_path)
-            if not self._zk_lock.acquire(blocking=False):
+            if not self._zk_lock.acquire(blocking=True, timeout=self._lock_timeout):
                 sys.exit(self._exitcode)
         else:
             raise RuntimeError("ZK flock enabled, but zookeeper is not configured")
--- a/ch_backup/logic/lock_manager.py
+++ b/ch_backup/logic/lock_manager.py
@@ -108,23 +108,13 @@ class LockManager:
             if not client.exists(self._process_zk_lockfile_path):
                 client.create(self._process_zk_lockfile_path, makepath=True)
             self._zk_lock = client.Lock(self._process_zk_lockfile_path)
-            force_exit = self._exitcode == 0
             try:
-                if not self._zk_lock.acquire(blocking=True, timeout=self._lock_timeout):
-                    fail_on_lock_acquiring(force_exit)
-            except LockTimeout as e:
-                fail_on_lock_acquiring(force_exit, e)
+                _ = self._zk_lock.acquire(blocking=True, timeout=self._lock_timeout)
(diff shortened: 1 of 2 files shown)


ch_backup/logic/lock_manager.py
```diff
@@ -41,6 +41,7 @@ def __init__(self, lock_conf: dict, zk_ctl: Optional[ZookeeperCTL]) -> None:
         self._exitcode = lock_conf.get("exitcode")
         self._distributed = zk_ctl is None
         self._disabled = False
+        self._lock_timeout = lock_conf.get("lock_timeout")
```
dstaroff, line 44: https://github.com/yandex/ch-backup/actions/runs/6326642673/job/17181029960?pr=73
You can add `too-many-instance-attributes` to the 21'st line to get rid of pylint error

dstaroff (commented):
The rest LGTM

dstaroff (approved):

Alex-Burmak:
Tests failed with `kazoo.exceptions.LockTimeout: Failed to acquire lock on /ch_backup/zk_flock_path after 60 seconds`.

dstaroff:
> Tests failed with `kazoo.exceptions.LockTimeout: Failed to acquire lock on /ch_backup/zk_flock_path after 60 seconds`.

Yes, we found it out, thank you. Maria is now working on proper solution for this

dstaroff (approved):

ch_backup/logic/lock_manager.py
```diff
@@ -106,7 +108,23 @@ def _zk_flock(self) -> None:
             if not client.exists(self._process_zk_lockfile_path):
                 client.create(self._process_zk_lockfile_path, makepath=True)
             self._zk_lock = client.Lock(self._process_zk_lockfile_path)
-            if not self._zk_lock.acquire(blocking=False):
-                sys.exit(self._exitcode)
+            force_exit = self._exitcode == 0
```
aalexfvk, line 111: Semantics of using `exitcode` config field was changed by this. Originally it was exit code in case of failed zk lock acquisition. But now only `0` will be actual exit code. 
As for exception, it can be catched and swallowed (maybe unintentionally).
Is it intended ?

ch_backup/logic/lock_manager.py
```diff
@@ -106,7 +108,23 @@ def _zk_flock(self) -> None:
...
+def fail_on_lock_acquiring(force_exit: bool, e: Optional[Exception] = None) -> None:
+    """
+    Determines how to exit from the lock acquisition process
+    """
+    if force_exit:
+        sys.exit(0)
+    err = RuntimeError("Lock was not acquired")
+    if e is not None:
+        raise err from e
+    raise err
```
aalexfvk, line 130: Does it handled somewhere ? How about 
- Just `sys.exit(self._exitcode)` here or
- Handling `lock_config.exitcode` where we catch and handle this exception to respect `lock_conf.exitcode` in all cases

ch_backup/logic/lock_manager.py
```diff
@@ -106,7 +108,23 @@ def _zk_flock(self) -> None:
             if not client.exists(self._process_zk_lockfile_path):
                 client.create(self._process_zk_lockfile_path, makepath=True)
             self._zk_lock = client.Lock(self._process_zk_lockfile_path)
-            if not self._zk_lock.acquire(blocking=False):
-                sys.exit(self._exitcode)
+            force_exit = self._exitcode == 0
+            try:
+                if not self._zk_lock.acquire(blocking=True, timeout=self._lock_timeout):
```
aalexfvk, line 113: As we are using `blocking=True`, `acquire()` raises an exception if the lock wasn't acquired within `timeout` seconds
https://github.com/python-zk/kazoo/blob/92bd0c2a8540697633564deed683f87d1f611bf5/kazoo/recipe/lock.py#L155
We shouldn't check the return value in that case.

ch_backup/logic/lock_manager.py
```diff
@@ -106,7 +108,17 @@ def _zk_flock(self) -> None:
             if not client.exists(self._process_zk_lockfile_path):
                 client.create(self._process_zk_lockfile_path, makepath=True)
             self._zk_lock = client.Lock(self._process_zk_lockfile_path)
-            if not self._zk_lock.acquire(blocking=False):
+            try:
+                lock_acquired = self._zk_lock.acquire(
```
aalexfvk, line 112: Do we need `lock_acquired` variable ? Why not just call exit in an `except` block ?

ch_backup/logic/lock_manager.py
```diff
@@ -106,7 +108,17 @@ def _zk_flock(self) -> None:
...
+                    blocking=True, timeout=self._lock_timeout
+                )
+            except LockTimeout:
+                lock_acquired = False
+                if self._exitcode != 0:
```
aalexfvk, line 117: I think it is helpful to print log unconditionally, to avoid silent exit/.

aalexfvk (changes requested):

aalexfvk (approved):

PR #79, opened 2023-10-16
Fix delete memory usage

MikhailBurdukov:
Do not keep full metadata for all backups in the same time.

### DIFF
--- a/ch_backup/ch_backup.py
+++ b/ch_backup/ch_backup.py
@@ -309,7 +309,7 @@ class ClickhouseBackup:
         retained_backups = []
         with self._context.locker():
             for i, backup in enumerate(
-                self._context.backup_layout.get_backups(use_light_meta=False)
+                self._context.backup_layout.get_backups(use_light_meta=True)
             ):
                 if backup.name == backup_name:
                     deleting_backups.append(backup)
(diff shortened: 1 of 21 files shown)


ch_backup/config.py
```diff
@@ -154,7 +154,7 @@ def _as_seconds(t: str) -> int:
     },
     "loguru": {
         "formatters": {
-            "ch-backup": "{time:YYYY-MM-DD H:m:s,SSS} {process.name:11} {process.id:5} [{level:8}] {extra[logger_name]}: {message}",
+            "ch-backup": "{time:YYYY-MM-DD HH:mm:ss,SSS} {process.name:11} {process.id:5} [{level:8}] {extra[logger_name]}: {message}",
```
Alex-Burmak, line 157: Looks unrelated to the memory usage fix.

aalexfvk: It looks like the branch was forked not from main, because this changes already in main.

ch_backup/ch_backup.py
```diff
@@ -309,7 +309,7 @@ def delete(
         retained_backups = []
         with self._context.locker():
             for i, backup in enumerate(
-                self._context.backup_layout.get_backups(use_light_meta=False)
+                self._context.backup_layout.get_backups(use_light_meta=True)
```
Alex-Burmak, line 312: Suggest to add a comment describing why it's important to use light metadata here. Something like `# Use light metadata in backups iteration to avoid high memory usage.`.

MikhailBurdukov:
Fixed review suggestions and made same changes for  `purge` command.

Alex-Burmak (approved):

aalexfvk (approved):

PR #81, opened 2023-10-24
Upload backup metadata during backup

aalexfvk:
Upload backup metadata with uploaded parts periodically during backup process. Interval of updating is set under `backup` config key `update_metadata_interval`. Default: 30 min.

Problem being solved:
`ch-backup` updates backup metadata in S3 storage after all parts for a table has been uploaded. If creation of backup is stopped unexpectedly (e.g. SIGKILL) all uploaded parts for a table can't be used for deduplication by next backups, because metadata about them was not uploaded.

### DIFF
--- a/ch_backup/logic/upload_part_observer.py
+++ b/ch_backup/logic/upload_part_observer.py
@@ -2,12 +2,8 @@
 import time
 from typing import List, Optional
 
-from humanfriendly import parse_timespan
-
 from ch_backup.backup.metadata.part_metadata import PartMetadata
 from ch_backup.backup_context import BackupContext
-
-DEFAULT_INTERVAL = parse_timespan("30 min")
 
 
 class UploadPartObserver:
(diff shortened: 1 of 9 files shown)


ch_backup/logic/upload_part_observer.py
```diff
@@ -0,0 +1,50 @@
+"""Uploading part observer."""
+import time
+from typing import List, Optional
+
+from humanfriendly import parse_timespan
+
+from ch_backup.backup.metadata.part_metadata import PartMetadata
+from ch_backup.backup_context import BackupContext
+
+DEFAULT_INTERVAL = parse_timespan("30 min")
```
Alex-Burmak, line 10: Looks redundant as there is a default in `config.py`.

Alex-Burmak (approved):

ch_backup/logic/upload_part_observer.py
aalexfvk: You are right. It was done due to one old unit test. But I have reworked it.

ch_backup/storage/async_pipeline/pipeline_executor.py
```diff
@@ -169,15 +170,30 @@ def wait(self, keep_going: bool = False) -> None:
...
         if is_async and self._exec_pool:
-            return self._exec_pool.submit(job_id, pipeline)
-
-        return pipeline()
+            return self._exec_pool.submit(job_id, pipeline, callback)
+
+        try:
+            result = pipeline()
+        except Exception as ex:
+            if callback:
```
MikhailBurdukov, line 190: When exception occurs then we will call a `callback(ex)`
But inside `class UploadPartObserver` we will just ignore it. Why? Maybe remove it?

PR #82, opened 2023-10-25
Thread safe retry for s3.

MikhailBurdukov:
Switched `retry` decorator from tenacity  to  own `retry_exponential` one.

### DIFF
--- a/ch_backup/storage/engine/s3/s3_retry.py
+++ b/ch_backup/storage/engine/s3/s3_retry.py
@@ -5,12 +5,13 @@ from abc import ABCMeta
 from functools import wraps
 from http.client import HTTPException
 from inspect import isfunction
-from typing import TYPE_CHECKING, Any, Callable, TypeVar
+from random import uniform
+from threading import Lock
+from time import sleep
+from typing import TYPE_CHECKING, Any, Callable, TypeVar, Union
 
 from botocore.exceptions import BotoCoreError, ClientError
 from urllib3.exceptions import HTTPError
-
-from ch_backup.util import retry
 
 if TYPE_CHECKING:
     from ch_backup.storage.engine import (
@@ -48,7 +49,9 @@ class S3RetryMeta(ABCMeta):
         """
 
         @wraps(func)
-        @retry(max_attempts=30, max_interval=180, exception_types=S3RetryingError)
+        @retry_exponential(
+            max_attempts=30, max_interval=180, exception_types=S3RetryingError
+        )
         def wrapper(self: "S3StorageEngine", *args: Any, **kwargs: Any) -> RT:
             try:
                 return func(self, *args, **kwargs)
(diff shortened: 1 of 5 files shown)


ch_backup/storage/engine/s3/s3_retry.py
```diff
@@ -57,3 +62,93 @@ def wrapper(self: "S3StorageEngine", *args: Any, **kwargs: Any) -> RT:
...
+
+        def calculate_sleep_time(self, retry_count):
+            """
+            Calculate time to sleep.
+            """
+
+            high_value = (2**retry_count) * multiplier
+            return uniform(0, min(high_value, max_interval))  # nosec
+
+        def call(self, fn, *args, **kwargs):
```
aalexfvk, line 109: Consider overriding  `__call__` method

ch_backup/storage/engine/s3/s3_retry.py
```diff
@@ -57,3 +62,93 @@ def wrapper(self: "S3StorageEngine", *args: Any, **kwargs: Any) -> RT:
...
+                            time_to_sleep,
+                            attempts,
+                            exc,
+                        )
+
+                    sleep_function(self.calculate_sleep_time(retry_count))
+
+    def wrap(f):
+        @wraps(f)
+        def wrapped_f(*args, **kw):
```
aalexfvk, line 149: `kwargs` is more common name

tests/unit/utils.py
```diff
@@ -177,3 +177,14 @@ def assert_equal(actual, expected):
         assert (
             False
         ), f"\n   expected: {expected}\n    but was: {actual}\ndifferences:\n{diff.pretty()}"
+
+
+class TimeMocker:
```
aalexfvk, line 182: Let's move it to the separate module `time_mocker.py`

aalexfvk (changes requested):

ch_backup/storage/engine/s3/s3_retry.py
```diff
@@ -57,3 +61,92 @@ def wrapper(self: "S3StorageEngine", *args: Any, **kwargs: Any) -> RT:
...
+        self.sleep_function = sleep_function
+        self.verbose = verbose
+
+    def calculate_sleep_time(self, attempt):
+        """
+        Calculate time to sleep.
+        """
+
+        high_value = (2**attempt) * self.multiplier
+        return uniform(0, min(high_value, self.max_interval))  # nosec
```
aalexfvk, line 118: Nevertheless, I think fixed interval backoff is more preferable here. Pay attention on [this](https://github.com/jd/tenacity/blob/310058274ed22a345e9c3917c97b4afd6363d5a5/tenacity/wait.py#L186) suggestion from tenacity. Clickhouse itself uses exponential fixed backoff intervals when requsting S3.

tests/unit/utils.py
aalexfvk: Forget to delete ?

ch_backup/storage/engine/s3/s3_retry.py
```diff
@@ -57,3 +61,92 @@ def wrapper(self: "S3StorageEngine", *args: Any, **kwargs: Any) -> RT:
...
+
+    def __call__(self, fn, *args, **kwargs):
```
aalexfvk, line 120: Write comment please that an instance of this class can be called from different threads and should be thread safe. For example, here we don't update state of the class (instance variables) and use only local ones.

aalexfvk (approved):

Alex-Burmak:
Test failures for `latest` ClickHouse versions don't relate to the changes from this PR. It relates to https://github.com/yandex/ch-backup/issues/84.

 Merging the PR.

Issue #84, opened 2023-11-03
ClickHouse 23.10 support


Alex-Burmak:
The following test scenario fails for ClickHouse 23.10: 

```
  Scenario: Backup table with S3 storage policy                   # tests/integration/features/backup_multiple_disks.feature:39
    Given default configuration                                   # tests/integration/steps/common.py:12
    And a working s3                                              # tests/integration/steps/s3.py:12
    And a working clickhouse on clickhouse01                      # tests/integration/steps/clickhouse.py:15
    And a working clickhouse on clickhouse02                      # tests/integration/steps/clickhouse.py:15
    Given we have executed queries on clickhouse01                # tests/integration/steps/clickhouse.py:55
      """
      CREATE DATABASE IF NOT EXISTS test_db;
      CREATE TABLE test_db.table_01 (
          CounterID UInt32,
          UserID UInt32
      )
      ENGINE = MergeTree()
      PARTITION BY CounterID % 3
      ORDER BY UserID
      SETTINGS storage_policy = 's3';
      
      INSERT INTO test_db.table_01 SELECT 0, number FROM system.numbers LIMIT 10;
      INSERT INTO test_db.table_01 SELECT 1, number FROM system.numbers LIMIT 10;
      """
    When we create clickhouse01 clickhouse backup                 # tests/integration/steps/ch_backup.py:26
    Then we got the following backups on clickhouse01             # tests/integration/steps/ch_backup.py:146
      | num | state   | data_count | link_count |
      | 0   | created | 2          | 0          |
    When we restore clickhouse backup #0 to clickhouse02          # tests/integration/steps/ch_backup.py:86
      """
      cloud_storage_source_bucket: 'cloud-storage-01'
      cloud_storage_source_path: 'data'
      """
      Assertion Failed: execution failed with code 1, out: "Traceback (most recent call last):
        File "/usr/local/bin/ch-backup", line 33, in <module>
          sys.exit(load_entry_point('ch-backup', 'console_scripts', 'ch-backup')())
        File "/usr/local/lib/python3.10/site-packages/click/core.py", line 1157, in __call__
          return self.main(*args, **kwargs)
        File "/usr/local/lib/python3.10/site-packages/click/core.py", line 1078, in main
          rv = self.invoke(ctx)
        File "/usr/local/lib/python3.10/site-packages/click/core.py", line 1688, in invoke
          return _process_result(sub_ctx.command.invoke(sub_ctx))
[...]

PR #85, opened 2023-11-03
Update ClickHouse versions

Alex-Burmak:


.github/workflows/workflow.yml
```diff
@@ -10,6 +10,7 @@ concurrency:
 
 env:
   NO_VENV: true
+  LATEST_VERSION: 23.9.3.12
```
Alex-Burmak, line 13: Should be replaced with `latest` when https://github.com/yandex/ch-backup/issues/84 get fixed.

PR #87, opened 2023-11-09
Improving logging for lock manager

filmonari:
Improving logging for lock manager by adding information about shard and replica name on lock acquisition. This would make possible to list the lock contenders on LockTimeout errors, thus helping the investigation of relative issues

### DIFF
--- a/ch_backup/ch_backup.py
+++ b/ch_backup/ch_backup.py
@@ -1,6 +1,7 @@
 """
 Clickhouse backup logic
 """
+import platform
 from collections import defaultdict
 from copy import copy
 from datetime import timedelta
@@ -17,6 +18,7 @@ from ch_backup.backup.metadata import BackupMetadata, BackupState, TableMetadata
 from ch_backup.backup.restore_context import RestoreContext
 from ch_backup.backup.sources import BackupSources
 from ch_backup.backup_context import BackupContext
+from ch_backup.clickhouse.client import ClickhouseError
 from ch_backup.clickhouse.config import ClickhouseConfig
 from ch_backup.clickhouse.control import ClickhouseCTL
 from ch_backup.clickhouse.models import Database
@@ -104,6 +106,22 @@ class ClickhouseBackup:
 
         return backups
 
+    def _lock_id(self, operation: str) -> Optional[str]:
+        lock_id = operation
+        # if the replica name part of the id starts with '$' it means
+        # we used `platform.node()`
+        missing_shard = "/UNKNOWN"
+        missing_replica = f"/${platform.node()}"
+        lock_id = operation
+        try:
+            shard, replica = self._context.ch_ctl.get_shard_and_replica_name()
+            lock_id += f"/{shard}" if shard else missing_shard
+            lock_id += f"/{replica}" if replica else missing_replica
+        except ClickhouseError:
+            logging.error("Error in forming lock_id.", exc_info=True)
+            lock_id = f"{lock_id}{missing_shard}{missing_replica}"
+        return lock_id
+
     def backup(
         self,
         sources: BackupSources,
@@ -162,7 +180,7 @@ class ClickhouseBackup:
 
         skip_lock = self._check_schema_only_backup_skip_lock(sources)
 
-        with self._context.locker(disabled=skip_lock):
+        with self._context.locker(disabled=skip_lock, lock_id=self._lock_id("BACKUP")):
             self._context.backup_layout.upload_backup_metadata(
                 self._context.backup_meta
             )
@@ -287,7 +305,9 @@ class ClickhouseBackup:
 
         logging.info("All required databases are present in backup.")
 
-        with self._context.locker(distributed=not sources.schema_only):
+        with self._context.locker(
+            distributed=not sources.schema_only, lock_id=self._lock_id("RESTORE")
+        ):
             self._restore(
                 sources=sources,
                 db_names=databases,
@@ -311,7 +331,7 @@ class ClickhouseBackup:
         found = False
         deleting_backups = []
         retained_backups = []
-        with self._context.locker():
+        with self._context.locker(lock_id=self._lock_id("DELETE")):
             # Use light metadata in backups iteration to avoid high memory usage.
             for i, backup in enumerate(
                 self._context.backup_layout.get_backups(use_light_meta=True)
@@ -363,7 +383,8 @@ class ClickhouseBackup:
         retained_backups: List[BackupMetadata] = []
         deleting_backups: List[BackupMetadata] = []
         backup_names = self._context.backup_layout.get_backup_names()
-        with self._context.locker():
+
+        with self._context.locker(lock_id=self._lock_id("PURGE")):
             # Use light metadata in backups iteration to avoid high memory usage.
             for backup in self._context.backup_layout.get_backups(use_light_meta=True):
                 if backup.name not in backup_names:
--- a/ch_backup/ch_backup.py
+++ b/ch_backup/ch_backup.py
@@ -1,7 +1,6 @@
 """
 Clickhouse backup logic
 """
-import platform
 from collections import defaultdict
 from copy import copy
 from datetime import timedelta
@@ -18,7 +17,6 @@ from ch_backup.backup.metadata import BackupMetadata, BackupState, TableMetadata
 from ch_backup.backup.restore_context import RestoreContext
 from ch_backup.backup.sources import BackupSources
 from ch_backup.backup_context import BackupContext
-from ch_backup.clickhouse.client import ClickhouseError
 from ch_backup.clickhouse.config import ClickhouseConfig
 from ch_backup.clickhouse.control import ClickhouseCTL
 from ch_backup.clickhouse.models import Database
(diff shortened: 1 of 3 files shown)


ch_backup/clickhouse/control.py
```diff
@@ -284,6 +284,17 @@
 """
 )
 
+GET_SHARD_AND_REPLICA_NAME = strip_query(
+    """
+    SELECT
+        last_value_respect_nullsArray(splitByString('/', zookeeper_path)) AS shard_name,
+        last_value_respect_nullsArray(splitByString('/', replica_path)) AS replica_name
```
filmonari, line 291: replica_name is also directly available in the table, but during tests is was always null. Thus I opted for splitting the replica_path...What it is better?

ch_backup/ch_backup.py
```diff
@@ -104,6 +106,22 @@ def list(self, state: BackupState = None) -> Sequence[BackupMetadata]:
 
         return backups
 
+    def _lock_id(self, operation: str) -> Optional[str]:
+        lock_id = operation
+        # if the replica name part of the id starts with '$' it means
+        # we used `platform.node()`
+        missing_shard = "/UNKNOWN"
+        missing_replica = f"/${platform.node()}"
```
filmonari, line 114: In case that getting replica and shard name through the query fails. It would be nice to still have some identifier for the locking node. Would `platform.node()` work for this purpose?

ch_backup/ch_backup.py
```diff
@@ -104,6 +106,22 @@ def list(self, state: BackupState = None) -> Sequence[BackupMetadata]:
...
+        lock_id = operation
+        try:
+            shard, replica = self._context.ch_ctl.get_shard_and_replica_name()
+            lock_id += f"/{shard}" if shard else missing_shard
+            lock_id += f"/{replica}" if replica else missing_replica
+        except ClickhouseError:
+            logging.error("Error in forming lock_id.", exc_info=True)
+            lock_id = f"{lock_id}{missing_shard}{missing_replica}"
```
MedvedewEM, line 122: It could be not easy to change lock_id string uniformly in future - because of "try" and "except" different cases.
What do you think about:
```
try:
    shard, replica = self._context.ch_ctl.get_shard_and_replica_name()
except ClickhouseError:
    logging.error(...)

shard = shard or "UNKNOWN"
replica = replica or f"${platform.node()}"
  
return f"{lock_id}{shard}{replica}"
```
?

ch_backup/logic/lock_manager.py
```diff
@@ -107,14 +122,18 @@ def _zk_flock(self) -> None:
...
             if not client.exists(self._process_zk_lockfile_path):
                 client.create(self._process_zk_lockfile_path, makepath=True)
-            self._zk_lock = client.Lock(self._process_zk_lockfile_path)
+            self._zk_lock = client.Lock(
+                self._process_zk_lockfile_path, identifier=self._lock_id
+            )
             try:
                 _ = self._zk_lock.acquire(blocking=True, timeout=self._lock_timeout)
-                logging.debug("Lock was acquired")
+                logging.debug("Lock was acquired.")
```
MedvedewEM, line 130: Lets clarify that it is ZK lock.
If we really need it maybe we should add the same log info to flock?

But for me it seems we do not need this at all - in case of problem with lock we are getting error's message.

filmonari: Yes, probably redundant. I will remove it

ch_backup/logic/lock_manager.py
```diff
@@ -44,18 +45,29 @@ def __init__(self, lock_conf: dict, zk_ctl: Optional[ZookeeperCTL]) -> None:
         self._disabled = False
         self._lock_timeout = lock_conf.get("lock_timeout")
 
-    def __call__(self, distributed: bool = True, disabled: bool = False):  # type: ignore
+    def __call__(self, distributed: bool = True, disabled: bool = False, lock_id: Optional[str] = None):  # type: ignore ## add 3rd parameter with log info (dict)
```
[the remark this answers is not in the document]
filmonari: yes, I forgot to remove it.

ch_backup/logic/lock_manager.py
filmonari: Do you think useful to include the zk_lock id in the error message?

ch_backup/clickhouse/control.py
aalexfvk: I'm not sure in this method. `zookeeper_path` can be the same for all shards and replicas. How about reading from `system.macros` ? As a rule it contains appropriate values for each replica. But this is not very reliable method.

@Alex-Burmak what do you think ?

ch_backup/logic/lock_manager.py
```diff
@@ -86,6 +98,9 @@ def __exit__(
             self._zk_lock.release()
             self._zk_client.__exit__(None, None, None)
 
+        logging.debug(self._log_msg_with_lock_id("Exiting lock."))
+        self._lock_id = None
```
aalexfvk, line 102: It looks suspicious. Where does this help, given that we always set `_lock_id` in the `__call__`. And now we can't do that:
```
locker = locker(lock_id="id")
with locker:
 # do smth
...
with locker:
 # do smth
```

ch_backup/logic/lock_manager.py
```diff
@@ -44,18 +45,29 @@ def __init__(self, lock_conf: dict, zk_ctl: Optional[ZookeeperCTL]) -> None:
...
         self._disabled = disabled
+        self._lock_id = lock_id
         return self
 
+    def _log_msg_with_lock_id(self, msg):
+        """
+        Append the lock id to a log message, if it is not null
+        """
+        if self._lock_id:
+            return f"{msg} Lock id is {self._lock_id}."
```
aalexfvk, line 62: We use loguru libarary in the project. Consider to use [opt()](https://loguru.readthedocs.io/en/stable/api/logger.html#loguru._logger.Logger.opt) facility and save new logger as an instance variable.

ch_backup/logic/lock_manager.py
filmonari: yes, you are right. It could be redundant. I just wanted to be sure that an lock id relative to an old operation was not left as _lock_id value. If we can consider this never happens, it makes thinks simpler.

ch_backup/clickhouse/control.py
Alex-Burmak: Agree. `system.replicas` is a wrong place for getting shard and replica names. It contains per table information. `zookeeper_path` and `replica_path` can be almost arbitrary values.

Vote for `system.macros`.

ch_backup/ch_backup.py
Alex-Burmak, line 122: Not sure that we need a fallback. It's better to choose one reliable approach for determining shard and replica names.

ch_backup/clickhouse/control.py
MikhailBurdukov: Can we use `replica_name` column directly from `system.replicas` instead of splitting `replica_path`? [docs](https://clickhouse.com/docs/ru/operations/system-tables/replicas)

ch_backup/logic/lock_manager.py
MikhailBurdukov: Agree, we can define new logger instance in config for such case.

filmonari: OK I'll try to create this new logger. What name should we use?

ch_backup/ch_backup.py
filmonari: Yes, that would be the best option. I think also that shard name could be neglected as knowing the replica (or hostname?) it is easy than to find the relative replica. What do you think guys? what would be the best approach? `platform.node()` only?

ch_backup/clickhouse/control.py
filmonari: thanks @Alex-Burmak. What would be best? `system.macros` or outside CH (see this [comment](https://github.com/yandex/ch-backup/pull/87/files#r1389337953))? what would a reliable way without CH to get this information?

ch_backup/ch_backup.py
Alex-Burmak: Agree, shard name can be omitted.

As for reliable way for getting hostname I suggest to use `socket.getfqdn()`. We already use it, for example, in https://github.com/yandex/ch-tools/blob/25728d2e3287c2e19d9c14259a6f466ddfe38c3a/ch_tools/common/clickhouse/client/clickhouse_client.py#L25

ch_backup/logic/lock_manager.py
filmonari: sorry I have missed this comment, and I have implemented a possible solution. I will keep it in a separate commit so it will be easy to revert if it is not good enough

ch_backup/logic/lock_manager.py
```diff
@@ -44,20 +47,24 @@ def __init__(self, lock_conf: dict, zk_ctl: Optional[ZookeeperCTL]) -> None:
...
 
-    def __call__(self, distributed: bool = True, disabled: bool = False):  # type: ignore
+    def __call__(self, distributed: bool = True, disabled: bool = False, operation: str = "UNKNOWN"):  # type: ignore
         """
         Call-method for context manager
         """
         self._distributed = distributed
         self._disabled = disabled
+        self._lock_id = f"{operation}/{socket.getfqdn()}"
+        self._logger = ChBackupLogger("lock-man", lock_id=self._lock_id)
```
aalexfvk, line 57: In my opinion it is over engeneering i would propose to just something like this. This will significantly reduce the PR.
Initially in my comment I confused `opt()` and `patch()`. Both methods return a custom logger.

```
def log_message(record):
    record['message'] = f"Changed message: {record['message']}"

self._logger = logger.patch(log_message)
self._logger.info('Hello')  # will output `Changed message: Hello``
```

PR #88, opened 2023-11-09
Add cli parameters to override config value or config section

MedvedewEM:
It merges default config + file-config + cli-config.
CLI-config has most priority.
And works like:
```
# ch-backup --config-value backup:deduplication_age_limit:days 150 list
...
final config after override and merge:  {'backup': {... 'deduplication_age_limit': {'days': 150} ... }}
...

# ch-backup --config-value backup '{"deduplication_age_limit": 150}' list
...
final config after override and merge:  {'backup': {... 'deduplication_age_limit': {'days': 150} ... }}
...

#ch-backup --config-value backup '{"deduplication_age_limit": 150, "retain_time": {'days': 150}}' --config-value 'backup:retain_time:days' 300 list
...
final config after override and merge:  {'backup': {... 'deduplication_age_limit': {'days': 150}, 'retain_time': {'days': 300} ... }}
...
```

### DIFF
--- a/ch_backup/cli.py
+++ b/ch_backup/cli.py
@@ -3,17 +3,15 @@
 Command-line interface.
 """
 import json
-import re
 import signal
 import sys
 import typing
 import uuid
-from collections import OrderedDict, defaultdict
+from collections import OrderedDict
 from functools import wraps
-from typing import Union
+from typing import Iterable, Tuple, Union
 
-from click import Choice, ParamType, Path, argument, pass_context, style
-from click.types import StringParamType
+from click import Choice, Path, argument, pass_context, style
 from cloup import (
     Color,
     Context,
@@ -25,7 +23,7 @@ from cloup import (
     option_group,
 )
 from cloup.constraints import constraint, mutually_exclusive
-from humanfriendly import InvalidTimespan, format_timespan, parse_timespan
+from humanfriendly import format_timespan
 from tabulate import tabulate
 
 from ch_backup.exceptions import TerminatingSignal
@@ -35,6 +33,7 @@ from .backup.metadata import BackupState, TableMetadata
 from .backup.sources import BackupSources
 from .ch_backup import ClickhouseBackup
 from .config import DEFAULT_CONFIG, Config
+from .params import JsonParamType, KeyValues, KeyValuesList, List, String, TimeSpan
 from .util import drop_privileges, setup_environment, utcnow
 from .version import get_version
 
@@ -100,6 +99,17 @@ signal.signal(signal.SIGINT, signal_handler)
     help="Disable certificate verification for https protocol.",
 )
 @option("--zk-hosts", type=str, help="Use specified zk hosts for connection to the ZK")
+@option(
+    "--config-parameter",
+    "config_parameters",
+    multiple=True,
+    type=(str, JsonParamType()),
+    metavar="PATH VALUE",
+    help="Paths and values to override ch-backup config values. "
+    'Path should contains a string with dot separated keys (e.g. "backup.min_interval.minutes"). '
+    "Value should be json-serializable string or plain value (string, true/false, number). "
+    "Can be specified multiple times to override several settings.",
+)
 @pass_context
 def cli(
     ctx: Context,
@@ -110,6 +120,7 @@ def cli(
     ca_path: Union[str, bool],
     insecure: bool,
     zk_hosts: str,
+    config_parameters: Iterable[Tuple[str, dict]],
 ) -> None:
     """Tool for managing ClickHouse backups."""
     if insecure:
@@ -126,6 +137,10 @@ def cli(
         cfg["clickhouse"]["ca_path"] = ca_path
     if zk_hosts is not None:
         cfg["zookeeper"]["hosts"] = zk_hosts
+
+    if config_parameters is not None:
+        cli_cfg = _build_cli_cfg_from_config_parameters(config_parameters)
+        cfg.merge(cli_cfg)
 
     logging.configure(cfg["loguru"])
     setup_environment(cfg["main"])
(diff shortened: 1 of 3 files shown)


ch_backup/cli.py
```diff
@@ -100,6 +122,17 @@ def signal_handler(signum, _frame):
...
 )
 @option("--zk-hosts", type=str, help="Use specified zk hosts for connection to the ZK")
+@option(
+    "--config-value",
+    "config_values",
+    multiple=True,
+    type=(str, JsonParamType()),
+    metavar="NAME VALUE",
+    help="Paths and values to override ch-backup config values. "
+    'Path should be a string separated with any of following symbols ".,:". '
```
Alex-Burmak, line 132: I think we can support only `.` as a separator for simplicity.

In dbaas tool several options are supported for compatibility with salt stack and postgresql syntaxes. Not sure that this feature is actual for ch-backup.

ch_backup/cli.py
```diff
@@ -100,6 +122,17 @@ def signal_handler(signum, _frame):
     help="Disable certificate verification for https protocol.",
 )
 @option("--zk-hosts", type=str, help="Use specified zk hosts for connection to the ZK")
+@option(
+    "--config-value",
```
Alex-Burmak, line 126: Suggest to change to  `"--parameter", "--config-parameter"`. I think it will be more intuitive clear cli option names.

Alex-Burmak (commented):
LGTM

Minor suggestions below.

ch_backup/cli.py
```diff
@@ -60,6 +60,28 @@ def signal_handler(signum, _frame):
...
+        try:
+            if re.fullmatch(
+                r'\s*([\[{"].*|true|false|null|\d+(\.\d+)?)\s*',
+                value,
+                re.MULTILINE | re.DOTALL,
+            ):
+                return json.loads(value)
+            return value.strip()
+        except json.JSONDecodeError as e:
+            raise ClickException(
```
aalexfvk, line 80: The `self.fail()` is [recommended](https://click.palletsprojects.com/en/8.1.x/parameters/#implementing-custom-types) way of signaling the conversion fails.

ch_backup/cli.py
```diff
@@ -60,6 +60,28 @@ def signal_handler(signum, _frame):
...
+
+    def convert(self, value, param, ctx):
+        try:
+            if re.fullmatch(
+                r'\s*([\[{"].*|true|false|null|\d+(\.\d+)?)\s*',
+                value,
+                re.MULTILINE | re.DOTALL,
+            ):
+                return json.loads(value)
+            return value.strip()
```
aalexfvk, line 78: How about returning dictionaries always (do parsing in that class). And just merge it (recursively) in the command handler if needed ? It looks like a better encapsulation.

ch_backup/cli.py
MedvedewEM: Ok, dot is enough, because I am mentioning it in "help" explicitly.
But actually it does not simplify code much :)

ch_backup/cli.py
MedvedewEM: I like "--config-parameter" more, because it's name is correlated with "config" parameter.

Btw, @Alex-Burmak, what do you think about breaking changes: remove parameters `--host/--protocol/...` from cli?
https://github.com/yandex/ch-backup/blob/main/ch_backup/cli.py#L89
Because now we can use current logic to define them.

Fix it now or take some time to test "--config-parameter"?

Alex-Burmak: I think it's OK to have the both new universal `--config-parameter` and existing specialized options. `--port`, `--protocol`, etc. are options for most common scenarios. `--config-parameter` option is for advanced configuration / scenarios.

ch_backup/cli.py
```diff
@@ -829,3 +867,42 @@ def _key_values_to_tables_metadata(
...
+
+
+def _build_cli_cfg_from_config_parameters(values: Iterable[Tuple[str, dict]]) -> dict:
+    """
+    Build config dict from specified keys and values in plain format.
+    Duplicate keys are ignored in favor of the first one.
+    """
+
+    def _split_key(key: str) -> typing.List[str]:
+        return re.split(r"[.]", key)
```
Alex-Burmak, line 879: Can be simplified to `key.split(".")`.

ch_backup/cli.py
MedvedewEM: But we are returning dictionary right now `return json.loads(value)` except when json is plain value/number/bool.

New parameter `--config-parameter` could be called multiple times at once.
And we can not _easily_ build one dict from several inputs here.

Do you mean create extended ParamType like MultipleJsonParamType which takes several inputs and do all logic from `_build_cli_cfg_from_config_parameters`?
Not sure about such work in this task :)
But for sure it will be easier to add same logic for other parameters in future.

ch_backup/cli.py
```diff
@@ -60,6 +60,28 @@ def signal_handler(signum, _frame):
 signal.signal(signal.SIGINT, signal_handler)
 
 
+class JsonParamType(ParamType):
```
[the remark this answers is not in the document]
MedvedewEM: Good notes, fixed all

Alex-Burmak (approved):

ch_backup/cli.py
aalexfvk: I also meant that when `JsonParamType` returns something that is not a dictionary it is a bit confusing. We could process one pair `--config-parameter full_path value`, and always return a dictionary like `{'path1': {'path2': value}}` in custom type. The `click` will process and put to a list several parameters itself.

And then call `config = merge(config_parameters)` in command handler to merge them all. 
But yes for this we need to move part of building logic to that class. 
As for me this is not a blocker, and if you think this is not rational, we can close this ticket.

MedvedewEM: Now I've got it, good idea in future.
It should works with both parameters - `full_path` and `value`, JsonParamType takes only one here.

So right now I will keep current behaviour.

PR #89, opened 2023-11-16
Made clickhouse-disks execution with separate config.

MikhailBurdukov:
Made `clickhouse-disks` execution with separate config which generates in run-time.

Resolves : #84

### DIFF
--- a/ch_backup/clickhouse/disks.py
+++ b/ch_backup/clickhouse/disks.py
@@ -2,7 +2,7 @@
 Clickhouse-disks controls temporary cloud storage disks management.
 """
 
-
+import copy
 import os
 from subprocess import PIPE, Popen
 from typing import Dict, List, Optional
@@ -59,16 +59,15 @@ class ClickHouseTemporaryDisks:
                 "Backup contains cloud storage data, cloud-storage-source-bucket must be set."
             )
 
-        self._disks: Dict[str, Disk] = {}
+        self._disks: Dict[str, Dict] = {}
         self._created_disks: Dict[str, Disk] = {}
 
     def __enter__(self):
-        self._disks = self._ch_ctl.get_disks()
+        self._disks = self._ch_config.config["storage_configuration"]["disks"]
         for disk_name in self._backup_meta.cloud_storage.disks:
-            disk = self._disks[disk_name]
             self._create_temporary_disk(
                 self._backup_meta,
-                disk,
+                disk_name,
                 self._source_bucket,
                 self._source_path,
                 self._source_endpoint,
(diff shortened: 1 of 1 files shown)


Alex-Burmak:
Предлагаю выставить LATEST_CLICKHOUSE_VERSION в latest в workflow.yml, чтобы проверить изменения для ClickHouse 23.10.

https://github.com/yandex/ch-backup/blob/a7bd5c6b9442b592d007eda77b29a001c57207d3/.github/workflows/workflow.yml#L13

ch_backup/clickhouse/disks.py
```diff
@@ -59,16 +59,15 @@ def __init__(
                 "Backup contains cloud storage data, cloud-storage-source-bucket must be set."
             )
 
-        self._disks: Dict[str, Disk] = {}
+        self._disks: Dict[str, Dict] = {}
         self._created_disks: Dict[str, Disk] = {}
 
     def __enter__(self):
-        self._disks = self._ch_ctl.get_disks()
+        self._disks = self._ch_config.config["storage_configuration"]["disks"]
```
aalexfvk, line 66: Why do you get rid of using `get_disks()` ?

MikhailBurdukov: Actually I just move it `copy_part` method. It only one place where we really need it.  To generate conf files in runtime we need to keep configuration of disks in this class. So `_disks` by now contains name of the disk and its configuration to provide ability to create configuration for `clickHouse-server`(ch_backup/clickhouse/disks.py:125) and  `clickhouse-disks`(ch_backup/clickhouse/disks.py:193).

ch_backup/clickhouse/disks.py
```diff
@@ -187,9 +190,14 @@ def _copy_dir(
                 to_path,
             ]
 
+        ch_disks_config_path = "/tmp/clickhouse-disks-config.xml"
+        self._render_disks_config(
+            ch_disks_config_path,
+            {from_disk: self._disks[from_disk], to_disk: self._disks[to_disk]},
+        )
```
Alex-Burmak, line 197: It seems we will re-generate the config file for each invocation of `_copy_dir`. Suggest to move to `ClickHouseTemporaryDisks.__enter__`.

MikhailBurdukov: Yes, it will re-generate but in `_render_disks_config` we will create config only for two disk: dest and source, instead of creating single config with all available disk. RN `clickhouse-disk` will load all disk from the config (https://github.com/ClickHouse/ClickHouse/issues/56791#issue-1994799242). Imo better to prevent unnecessary storage creating.

ch_backup/clickhouse/disks.py
```diff
@@ -87,63 +98,66 @@ def __exit__(self, exc_type, *args, **kwargs):
...
-                        },
-                    }
-                },
-                f,
-                pretty=True,
-            )
+        self._render_disks_config(
+            _get_config_path(self._config_dir, tmp_disk_name),
+            {tmp_disk_name: disk_config},
+        )
```
Alex-Burmak, line 140: Do we still need this file? Isn't tmp disk needed only for clickhouse-disks?

MikhailBurdukov: It is a config for  clickhouse-server, we need it to do not create tmp_disk manually (like data_path and downloading metadata)

Alex-Burmak (approved):

aalexfvk (approved):

PR #93, opened 2023-11-30
Enable zookeeper clean by default for restore cmd.

MikhailBurdukov:


ch_backup/cli.py
```diff
@@ -476,7 +476,7 @@ def backup_command(
 @option_group(
     "ZooKeeper",
     option(
-        "--clean-zookeeper",
+        "--no-clean-zookeeper",
```
Alex-Burmak, line 479: Suggest to replace with:
```
"--clean-zookeeper/--no-clean-zookeeper",
default=True,
```

PR #109, opened 2024-03-17
Remove loading all files in memory at once

kirillgarbar:
Replace loading all file names in memory from `shadow` directory in `upload_cloud_storage_metadata` and all frozen parts in `_backup_frozen_table_data`  with `os.scandir` method.

In other places such as [upload_access_control_files](https://github.com/kirillgarbar/ch-backup/blob/ff08914bf490c9d7221787be36753ce9231cb98a/ch_backup/backup/layout.py#L122) and [here](https://github.com/kirillgarbar/ch-backup/blob/ff08914bf490c9d7221787be36753ce9231cb98a/ch_backup/clickhouse/control.py#L657) file lists should not be large and it may be tricky to not load them in memory as usual.


ch_backup/util.py
```diff
@@ -87,6 +89,51 @@ def list_dir_files(dir_path: str) -> List[str]:
...
 
+def scan_dir_files(
+    dir_path: Path, exclude_file_names: Optional[List[str]] = None
+) -> Iterable[str]:
+    """
+    Yields relative file paths in a given directory and excludes files with given names
+    """
+
+    def scan_recursive(dir_path: Path, relative_prefix: Path = None) -> Iterable[str]:
+        with os.scandir(dir_path) as scan:
```
aalexfvk, line 100: It's ok, but why not `os.walk()` ? This approach is just harder to maintain.

tests/unit/test_calculators.py
```diff
@@ -69,7 +73,23 @@ def test_calc_aligned_file_size(
 def test_calc_tarball_size(
     name_lens: List[int], data_size: int, expected_size: int
 ) -> None:
-    names = ["0" * name_len for name_len in name_lens]
+    try:
+        # Assuming name_lens size is less than 10
+        names = [f"{i}" * name_len for i, name_len in enumerate(name_lens)]
+        cwd = Path(__file__).parent.resolve()
+        test_dir_path = cwd / "test"
+        os.mkdir(test_dir_path)
```
aalexfvk, line 81: Consider using of [TemporaryDirectory](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory)

tests/unit/test_calculators.py
```diff
@@ -69,7 +73,23 @@ def test_calc_aligned_file_size(
...
+        for name in names:
+            with open(test_dir_path / name, "w", encoding="utf-8") as _:
+                continue
+        assert calc_tarball_size_scan(test_dir_path, data_size) == expected_size
+    except OSError as e:
+        # open() will raise file name too long in some cases
+        if e.errno != 36:
```
aalexfvk, line 88: `errno.ENAMETOOLONG`

ch_backup/util.py
kirillgarbar: `os.walk` yields all files and directory names from a given directory at once, so for a directory with a lot of files or other directories returned tuple might be large

tests/unit/test_calculators.py
```diff
@@ -73,24 +73,20 @@ def test_calc_aligned_file_size(
...
-        if e.errno != 36:
-            raise
-    finally:
-        if test_dir_path.is_dir():
-            shutil.rmtree(test_dir_path)
-
+    # Assuming name_lens size is less than 10
+    names = [f"{i}" * name_len for i, name_len in enumerate(name_lens)]
+    cwd = Path(__file__).parent.resolve()
+    with TemporaryDirectory(prefix=str(cwd) + "/") as dir_name:
```
aalexfvk, line 79: It seems you misuse `prefix` here. If you want to define a parent directory you need to use `dir=` parameter. But I would propose not to use neither `prefix` nor `dir`.

PR #114, opened 2024-03-26
Add fail reason in backup metadata

kirillgarbar:
Saving exception type and message for using them in monitoring

### DIFF
--- a/ch_backup/backup/metadata/backup_metadata.py
+++ b/ch_backup/backup/metadata/backup_metadata.py
@@ -64,7 +64,7 @@ class BackupMetadata:
         self.cloud_storage: CloudStorageMetadata = CloudStorageMetadata()
 
         self._state = BackupState.CREATING
-        self._fail_reason: Optional[str] = None
+        self._exception: Optional[str] = None
         self._databases: Dict[str, dict] = {}
         self._access_control = AccessControlMetadata()
         self._user_defined_functions: List[str] = []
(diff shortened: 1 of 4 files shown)


tests/integration/features/metadata.feature
```diff
@@ -42,3 +42,25 @@ Feature: Metadata
...
+    """
+    Given we have executed command on clickhouse01
+    """
+    rm -rf /var/lib/clickhouse/data/default/foo/all_1_1_0
+    """
+    When we try to create clickhouse01 clickhouse backup
+    Then metadata of clickhouse01 backup #0 contains value for "fail_reason" which begins with
+    """
+    FileNotFoundError: [Errno 2] No such file or directory
+    """
```
Alex-Burmak, line 66: I'd expect that the field named `fail_reason` contains more high-level values: "disk quota exceeded", "memory limit exceeded", "internal error", etc.

If we just add an exception to metadata, suggest to rename the field to `exception`.

ch_backup/backup/metadata/backup_metadata.py
```diff
@@ -178,6 +191,7 @@ def load(cls, data: dict) -> "BackupMetadata":
             backup.size = meta["bytes"]
             backup.real_size = meta["real_bytes"]
             backup._state = BackupState(meta["state"])
+            backup._fail_reason = meta["fail_reason"] if "fail_reason" in meta else None
```
Alex-Burmak, line 194: Suggest to replace `meta["fail_reason"] if "fail_reason" in meta else None` with `meta.get("fail_reason", None)` for simplicity and consistency with the code below.

Alex-Burmak (approved):

PR #151, opened 2024-05-07
Disable bulk delete for files that break XML

kirillgarbar:
The range of characters in regexp corresponds to almost all ordinary characters, bulk-delete seems not to have any problems with them.
Also I decided to try always delete in bulk first in case the character is not in the range and does not break the deletion.

### DIFF
--- a/ch_backup/storage/engine/s3/s3_engine.py
+++ b/ch_backup/storage/engine/s3/s3_engine.py
@@ -18,7 +18,6 @@ from ch_backup.storage.engine.s3.s3_client_factory import (
 from ch_backup.storage.engine.s3.s3_multipart_uploader import S3MultipartUploader
 from ch_backup.storage.engine.s3.s3_retry import S3RetryMeta
 from ch_backup.type_hints.boto3.s3 import S3Client
-from ch_backup.util import contains_unusual_characters
 
 
 class S3StorageEngine(PipeLineCompatibleStorageEngine, metaclass=S3RetryMeta):
@@ -88,12 +87,16 @@ class S3StorageEngine(PipeLineCompatibleStorageEngine, metaclass=S3RetryMeta):
         """
         Delete multiple files from S3
         """
-        # https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html#S3.Client.delete_objects
-        if not self._bulk_delete_enabled:
+
+        def delete_by_one(remote_paths: Sequence[str]) -> None:
             for remote_path in remote_paths:
                 self.delete_file(remote_path)
+
+        if not self._bulk_delete_enabled:
+            delete_by_one(remote_paths)
             return
 
+        # https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html#S3.Client.delete_objects
         try:
             objects_to_delete: list = [
                 {"Key": path.lstrip("/")} for path in remote_paths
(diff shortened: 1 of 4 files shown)


ch_backup/storage/engine/s3/s3_engine.py
```diff
@@ -88,15 +89,36 @@ def delete_files(self, remote_paths: Sequence[str]) -> None:
...
+                else:
+                    objects_to_delete_bulk.append({"Key": path.lstrip("/")})
+
+            if objects_to_delete_bulk:
+                self._s3_client.delete_objects(
+                    Bucket=self._s3_bucket_name,
+                    Delete={"Objects": objects_to_delete_bulk},
+                )
+            for remote_path in objects_to_delete_single:
                 self.delete_file(remote_path)
```
Alex-Burmak, line 122: I think fallback logic can be simplified. If bulk delete fails with `MalformedXML` then simply delete all files one by one.

ch_backup/util.py
```diff
@@ -41,6 +41,7 @@
     timedelta(seconds=-1 * (time.altzone if time.daylight else time.timezone))
 )
 _ALLOWED_NAME_CHARS = set(["_"] + list(ascii_letters) + list(digits))
+_UNUSUAL_CHARS_REGEX = re.compile(r"[^\u0020-\u007e]")
```
MikhailBurdukov, line 44: After the [comment](https://github.com/yandex/ch-backup/pull/151/files#r1594025940), im think we don't need it. But if you will keep it:

Actually it isn't clear for me what does this mean.  Do we really need to use unicode here? Why not ASCII? As I can see it is just ASCII Punctuation & Symbols chars. See (https://en.wikipedia.org/wiki/List_of_Unicode_characters)

I think it is better to use simple for loop with  `string.punctuation`. Code will be more readable. Can we do it like that?

ch_backup/storage/engine/s3/s3_engine.py
```diff
@@ -88,15 +89,36 @@ def delete_files(self, remote_paths: Sequence[str]) -> None:
...
+            objects_to_delete_single: list = []
+            objects_to_delete_bulk: list = []
+            for path in remote_paths:
+                if contains_unusual_characters(path):
+                    objects_to_delete_single.append(path)
+                else:
+                    objects_to_delete_bulk.append({"Key": path.lstrip("/")})
+
+            if objects_to_delete_bulk:
+                self._s3_client.delete_objects(
```
MikhailBurdukov, line 117: I think it is not related, but...:
 From doc: [The request can contain a list of up to 1000 keys that you want to delete.](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3/client/delete_objects.html)
Can we limit size of the objects list in the single `delete_objects` call?

Alex-Burmak: Should be already handled by `DeleteMultipleStorageStage` from https://github.com/yandex/ch-backup/blob/master/ch_backup/storage/stages/storage.py

tests/integration/features/backup_delete.feature
```diff
@@ -110,6 +110,45 @@ Feature: Backup & Clean & Restore
...
+    """
+    echo -e 'CREATE DATABASE unusual_char;' | clickhouse-client
+    """
+    And we execute command on clickhouse01
+    """
+    echo -e 'CREATE TABLE unusual_char."tab\x07l./-_e" (n UInt64) ENGINE = MergeTree() ORDER BY n;' | clickhouse-client
+    """
+    And we execute command on clickhouse01
+    """
+    echo -e 'INSERT INTO unusual_char."tab\x07l./-_e" VALUES (1234);' | clickhouse-client
```
MikhailBurdukov, line 124: Can we use `execute query` or `clickhouse-client --query` or it is impossible because of unusual chars?

MikhailBurdukov (approved):

Alex-Burmak (approved):

PR #152, opened 2024-05-16
Use custom built-in python with ch-backup

ekopanev:
Added extra files for building ch-backup with built-in custom python version, currently it's compiling during the build (but later can be moved to a separate image/static archive. etc)

To build it with custom python we can run the following:
```
cd ch-backup
make build-python
./build_deb_with_python_in_docker.sh
```
with new env variables `PYTHON_BUILD_NAME` (eg: `bionic-python-amd64`) and `TARGET_PYTHON_VERSION` (eg: `3.12.3`)

It will build and compile target python version with target image (OS and arch), and then copy the whole python directory to the end deb package and setup it's usage (to not overlap with system python and be able to install all ch-backup dependencies into temporary venv)

There were a couple of issues with current `boto` dependencies and python 3.12.3, that's why I've split them by `python_version` (see also https://github.com/yandex/ch-backup/pull/147). Currently it works well with S3, with python 3.12.3 and upgraded boto libs:
```
boto3              1.20.54
botocore           1.23.54
```


debian/rules
```diff
@@ -7,6 +7,11 @@ PYTHON_MINOR := $(shell $(PYTHON) -c 'import sys; print(sys.version_info[1])')
 PYTHON_FROM := $(PYTHON_MAJOR).$(PYTHON_MINOR)
 PYTHON_TO := $(PYTHON_MAJOR).$(shell echo $$(( $(PYTHON_MINOR) + 1 )))
 
+# In case of using built-in python version (which is expected to have at least 3.12)
+# we don't care about this dependency and it can be removed at all later.
+ifeq ($(PYTHON_MINOR),12)
```
ekopanev, line 12: not sure how to avoid this dependency more elegant than that, maybe we can just set `SUBSTVARS` to an empty string for example

with built-in python we don't need this (as well as `Build-Depends`) - but we can clean it up later somehow

Makefile
```diff
@@ -211,6 +210,21 @@ check-environment:
 	touch .install-deps
 
 
+.PHONY: build-python
+build-python:
```
Alex-Burmak, line 214: Please add description of new target to the help (to `help` target below).

Dockerfile-deb-python-build
```diff
@@ -0,0 +1,52 @@
...
+        # Misc
+        curl \
+        locales \
+        # For building PyNacl library
+        libffi-dev libssl-dev libboost-all-dev libsodium-dev \
+    # Configure locales
+    && locale-gen en_US.UTF-8 \
+    && update-locale LANG=en_US.UTF-8 \
+    # Ensure that `python` refers to `python3` so that poetry works.
+    # It makes sense for ubuntu:18.04
```
Alex-Burmak, line 37: Is this comment correct? It refers to Ubuntu 18.04 but this Docerfile is based on Ubuntu 22.04 .

Alex-Burmak (commented):
LGTM except a few comments.

Also consider adding test job to GitHub CI with new feature.

debian/rules
Alex-Burmak: I suppose we can workaround lack of numeric comparison in debin rules using `shell` directive.

ekopanev: changed for using shell and proper comparison I guess

while this currently is needed for all builds with custom python, we can assume that we'll use it only for 3.12+, otherwise need to think how to remove this dependency properly

Dockerfile-deb-python-build
ekopanev: Partly, but I've changed it a bit. It was supposed to use the same OS for building python as well as for building ch-backup later, so we currently use proper OS via arg

Alex-Burmak (approved):

aalexfvk:
Do we actually need custom building instead of [official](https://hub.docker.com/_/python/tags) images or installing from [ppa](https://launchpad.net/~deadsnakes/+archive/ubuntu/ppa) ? Did you consider using [venv --copies](https://docs.python.org/3/library/venv.html#creating-virtual-environments)

debian/rules
aalexfvk: Nevertheless, I think that checking environment variable here if embedded python is used and using default value `SUBSTVARS := -Vpython:Depends=""` are more clear

Dockerfile-deb-python-build
```diff
@@ -0,0 +1,51 @@
+ARG PYTHON_BUILD_NAME
+ARG BASE_IMAGE=ubuntu:22.04
+
+FROM --platform=$TARGETPLATFORM $PYTHON_BUILD_NAME-build:latest as pybuild
```
aalexfvk, line 4: How about passing full image name here without assumption `*-build:latest`

Dockerfile-python-build
```diff
@@ -0,0 +1,43 @@
...
+        libreadline-dev \
+        libsqlite3-dev \
+        libssl-dev \
+        wget \
+        xz-utils \
+        zlib1g-dev \
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/*
+
+ARG PYTHON_INSTALL_PREFIX=/src/python
```
aalexfvk, line 25: I would propose to remove default value and add checking if it is empty. Like so
```
ARG PYTHON_INSTALL_PREFIX
RUN test -n "$PYTHON_INSTALL_PREFIX" || (echo "PYTHON_INSTALL_PREFIX must be provided" && false)
```

Dockerfile-deb-python-build
```diff
@@ -0,0 +1,51 @@
...
+
+# Project directory must be mounted here
+VOLUME /src
+WORKDIR /src
+
+# For compiling PyNACL library, which is used by ch-backup
+# See https://pynacl.readthedocs.io/en/latest/install/#linux-source-build
+ENV SODIUM_INSTALL=system
+
+ENV PYTHON=/opt/yandex/ch-backup/python/bin/python3.12
```
aalexfvk, line 48: Let's remove hard-coded versions where it is possible

Dockerfile-python-build
```diff
@@ -0,0 +1,43 @@
...
+RUN mkdir -p $PYTHON_INSTALL_PREFIX
+
+ARG PYTHON_VERSION=3.12.3
```
aalexfvk, line 28: Let's move all ARG at the beginning of dockerfile

.github/workflows/workflow.yml
```diff
@@ -82,3 +82,27 @@ jobs:
...
+    - name: Build with custom python
+      shell: bash
+      env:
+        PYTHON_BUILD_NAME: ${{ matrix.distro }}-python-${{ matrix.arch }}
+        TARGET_PYTHON_VERSION: 3.12.3
+        DEB_TARGET_PLATFORM: linux/${{ matrix.arch }}
+        DEB_BUILD_DISTRIBUTION: ubuntu:${{ matrix.distro }}
+        DOCKER_BUILDKIT: 1
+      run: |
+        make build-python
```
aalexfvk, line 108: How about building full deb here and testing if it is workable like we do [here](https://github.com/yandex/ch-tools/blob/6e513f6b1bbe7e53841daff7823e830a868e29d7/.github/workflows/workflow.yml#L142)

aalexfvk (changes requested):

ekopanev:
Support of py3.12 was added instead within https://github.com/yandex/ch-backup/pull/167, so will close this one for now

PR #153, opened 2024-05-21
Use __slots__ for frequently created classes

kirillgarbar:

PR #155, opened 2024-05-24
MDB-21949 Use tmp folder for restore/backup access control entities

MikhailBurdukov:


tests/integration/features/access_control_replicated_backup.feature
```diff
@@ -45,14 +45,7 @@ Feature: Backup and restore functionality of replicated access control entities
     Then we got the following backups on clickhouse01
       | num | state   | data_count | link_count |
       | 0   | created | 1          | 0          |
-    When we execute command on clickhouse01
```
MikhailBurdukov, line 48: Not needed anymore, because we don't copy entities from the ZK to `/var/lib/clickhouse/access`

MikhailBurdukov:
Fix for tests : #153

ch_backup/backup/layout.py
```diff
@@ -120,19 +120,18 @@ def upload_access_control_file(self, backup_name: str, file_name: str) -> None:
             raise StorageError(msg) from e
 
     def upload_access_control_files(
-        self, backup_name: str, file_names: List[str]
+        self, path: str, backup_name: str, local_path: str, file_names: List[str]
```
aalexfvk, line 123: Maybe rename path to local_path ?

ch_backup/logic/access.py
```diff
@@ -28,13 +33,39 @@ def backup(self, context: BackupContext) -> None:
...
-            self._backup_replicated(access_control.acl_ids, context)
-        self._backup_local(access_control.acl_ids, context)
+        clickhouse_access_path = context.ch_ctl_conf["access_control_path"]
+        backup_tmp_path = os.path.join(
+            context.ch_ctl_conf["tmp_path"], context.backup_meta.name
+        )
+        user = context.ch_ctl_conf["user"]
+        group = context.ch_ctl_conf["group"]
+
+        create_dir(clickhouse_access_path, user, group)
```
aalexfvk, line 44: 1. Actually, I don't see a necessity of creating separate function for this (creating folder and chown it). Two separate commands are relatively expressive.
2. `clickhouse_access_path` does not exist here ?

ch_backup/logic/access.py
```diff
@@ -28,13 +33,39 @@ def backup(self, context: BackupContext) -> None:
...
+        with temporary_directory(backup_tmp_path, user, group):
+            objects = context.ch_ctl.get_access_control_objects()
+            context.backup_meta.set_access_control(objects)
+            access_control = context.backup_meta.access_control
+
+            if self._has_replicated_access(context):
+                self._backup_replicated(
+                    backup_tmp_path, access_control.acl_ids, context
+                )
+            self._backup_local(clickhouse_access_path, backup_tmp_path)
```
ekopanev, line 54: hm, am I right that now `_backup_replicated()` just copies data from ZK to tmp dir, while `_backup_local()` copies local files from access path to tmp and that's it?

And so:
- `_backup_local` will always replace (in case of replicated storage) actual data from ZK with local version (because currently we do not clean them up)
- we still will have a local (and soon "outdated") version of ACL in access folder anyway?

ch_backup/logic/access.py
```diff
@@ -48,11 +79,35 @@ def restore(self, context: BackupContext) -> None:
...
+        clickhouse_access_path = context.ch_ctl_conf["access_control_path"]
+        restore_tmp_path = os.path.join(
+            context.ch_ctl_conf["tmp_path"], context.backup_meta.name
+        )
+        user = context.ch_ctl_conf["user"]
+        group = context.ch_ctl_conf["group"]
+
+        create_dir(clickhouse_access_path, user, group)
+        with temporary_directory(restore_tmp_path, user, group):
+            if (
```
aalexfvk, line 92: Is this code needed in case of `has_replicated_access=True` ?

tests/integration/features/access_control_replicated_backup.feature
ekopanev: also, if we don't want to clean up access/tmp folders - should we change this step for checking tmp dir instead?

ch_backup/logic/access.py
MikhailBurdukov: > clickhouse_access_path does not exist here ?

Yes, clickhouse lazily creates `/var/lib/clickhouse/access`

ch_backup/logic/access.py
MikhailBurdukov: Which one? This?
https://github.com/yandex/ch-backup/blob/5dfdf661279defd79cbe35536961dee661792ed1/ch_backup/logic/access.py#L92-L95

Or
https://github.com/yandex/ch-backup/blob/5dfdf661279defd79cbe35536961dee661792ed1/ch_backup/logic/access.py#L90

ch_backup/logic/access.py
MikhailBurdukov: >clickhouse_access_path does not exist here ?

Yes, clickhouse lazily creates `/var/lib/clickhouse/access`

ch_backup/logic/access.py
```diff
@@ -28,13 +33,39 @@ def backup(self, context: BackupContext) -> None:
...
-        self._backup_local(access_control.acl_ids, context)
+        clickhouse_access_path = context.ch_ctl_conf["access_control_path"]
+        backup_tmp_path = os.path.join(
+            context.ch_ctl_conf["tmp_path"], context.backup_meta.name
+        )
+        user = context.ch_ctl_conf["user"]
+        group = context.ch_ctl_conf["group"]
+
+        create_dir(clickhouse_access_path, user, group)
+        with temporary_directory(backup_tmp_path, user, group):
```
aalexfvk, line 45: 1. Do we need to invent own class instead of ['TemporaryDirectory'](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory)
2. Does this class produce leftovers in case of exception and is it expected

ch_backup/logic/access.py
aalexfvk: https://github.com/yandex/ch-backup/blob/5dfdf661279defd79cbe35536961dee661792ed1/ch_backup/logic/access.py#L92-L95

ch_backup/logic/access.py
MikhailBurdukov: For TemporaryDirectory there are no option to keep folder in case of exceptions.
https://github.com/python/cpython/blob/59630f92d8223f80993e3646b0f734d27f4b8dd4/Lib/tempfile.py#L944-L966

ch_backup/logic/access.py
ekopanev: > Yes, we will have it. I think make sense to cleanup /var/lib/clickhouse/access in restore operation.

because I thought that originally we could have potential issue with local copies after each backup - in case of troubles with ZK we can't use replicated storage - so we can have a conflict with local sql files

so maybe we should clean it up after backup operation as well?

ch_backup/logic/access.py
```diff
@@ -48,11 +79,35 @@ def restore(self, context: BackupContext) -> None:
...
+                == BackupStorageFormat.TAR
+            ):
+                context.backup_layout.download_access_control(
+                    restore_tmp_path, context.backup_meta.name
+                )
+            else:
+                for name in _get_access_control_files(acl_ids):
+                    context.backup_layout.download_access_control_file(
+                        restore_tmp_path, context.backup_meta.name, name
+                    )
```
aalexfvk, line 103: Let's extract this to a separate function `_download_access_control`

ch_backup/logic/access.py
MikhailBurdukov: >so maybe we should clean it up after backup operation as well?

Im not sure because suppose:
1) we have 2 acl entries: 1 - old one without zk, 2 - with zk
2) perform backup -> removing from `/var/lib/clickhouse/acces`.
3) So after  ch-server restart we will lose 1 entry.

It seems fine to do cleanup on restore, but we will lose non-replicated records when we do a backup.

ch_backup/logic/access.py
aalexfvk: Why do we want to leave data in a temporary directory in case of exception ? It might be corrupted

MikhailBurdukov: Because it might be useful in trouble shooting. If we want to re-run operation(backup/restore), then old files will be replaced with a new one. Also it make sense to cleanup tmp folder before operation, if it exist.

aalexfvk: Ok, nevertheless this class is too specific for a general class(to be placed in `ch_backup/util.py`). Someone might want to accidentally reuse it for their own purposes
I see the following options:
1) Write a comment that this class is intended for using for access control backup procedure and write this motivation about keeping data.
2) Rename it to smth like `access_control_temp_directory`.
3) Create a fully functional replacement for `tempfile.TemporaryDirectory` with `keep` parameter. 
Simplified example
```
@contextmanager
def temporary_directory(keep=False):
    if keep:
        tmpdir = mkdtemp()
        print("Keeping tempdir: {}".format(tmpdir))
        yield tmpdir
    else:
        with TemporaryDirectory() as tmpdir:
            yield tmpdir
```

ch_backup/logic/access.py
```diff
@@ -48,11 +79,29 @@ def restore(self, context: BackupContext) -> None:
...
+            context.ch_ctl_conf["tmp_path"], context.backup_meta.name
+        )
+        user = context.ch_ctl_conf["user"]
+        group = context.ch_ctl_conf["group"]
+
+        if os.path.exists(clickhouse_access_path):
+            shutil.rmtree(clickhouse_access_path)
+
+        os.makedirs(clickhouse_access_path)
+        shutil.chown(clickhouse_access_path, user, group)
```
aalexfvk, line 94: It makes sense to extract this lines (except related tmp path) to a separate function with `clean` parameter.

aalexfvk (approved):

MikhailBurdukov:
@ekopanev any comments?

ch_backup/logic/access.py
ekopanev: > we have 2 acl entries: 1 - old one without zk, 2 - with zk

is it already possible to have 2 entities in different storages (local and replicated)? because I've tested it initially (22.x), and found that you can use only one storage type for that (if already no - maybe it makes sense)

and if 1 possible storage is still relevant - it's better to clean it up after a backup as well (think about your current state on the clusters - they all have local files), or move/remove them during migration from replicated to local

otherwise I think current approach works only for new clusters and with new ch-backup version

ch_backup/util.py
```diff
@@ -444,3 +444,51 @@ def replace_macros(string: str, macros: dict) -> str:
         pattern=r"{([^{}]+)}",
         repl=lambda m: macros.get(m.group(1), m.group(0)),
     )
+
+
+def chown_path(path: str, user: str, group: str) -> None:
+    """
+    Interface for chown
+    """
+    shutil.chown(path, user, group)
```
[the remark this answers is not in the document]
ekopanev: seems like unused function and can be deleted?

ekopanev (approved):
the rest LGTM

ekopanev (approved):

PR #158, opened 2024-05-29
Use ijson for backup metadata in deduplication

kirillgarbar:
Previously, all backup metadata was loaded as json. Now we load data for one database at a time.
`gc.collect` is added just in case after each database is processed, but maybe it is not necessary now.


aalexfvk:
Decided not to use approach with ijson in favour of https://github.com/yandex/ch-backup/pull/159

PR #159, opened 2024-06-05
Deduplicate with ClickHouse table

kirillgarbar:
DedupInfo and DatabaseDedupInfo classes are removed, DedupReferences is used to collect dedup info instead.
Dedup info is collected in clickhouse table, during deduplication another table is created and deduplication is done with JOIN by name and checksum query.
Unit tests are removed due to the need to use running clickhouse server.


ch_backup/clickhouse/control.py
```diff
@@ -170,7 +170,52 @@
...
+"""
+)
+CREATE_DEDUP_TABLE_CURRENT_SQL = strip_query(
+    """
+    CREATE TABLE {system_db}._deduplication_info_current (
+        name String,
+        checksum String,
+    )
+    ENGINE = MergeTree()
+    ORDER BY name
```
aalexfvk, line 204: And here let's add `checksum`

ch_backup/backup/deduplication.py
```diff
@@ -222,60 +206,65 @@ def _populate_dedup_info(
                         verified=verified,
                     )
 
+                    table_dedup_info.add(part.name)
```
aalexfvk, line 209: Where part names from `table_dedup_info` are used ?

ch_backup/backup/deduplication.py
```diff
@@ -49,69 +59,41 @@ def __init__(
...
-        """
-        return self._databases[database_name]
+DedupReferences = DefaultDict[str, DatabaseDedupReferences]
 
-    def __repr__(self):
-        return f"DedupInfo({dict(self._databases)})"
 
-    def __eq__(self, other):
-        return self.__dict__ == other.__dict__
+def _create_dedup_references() -> DedupReferences:
```
aalexfvk, line 77: `_create_empty_dedup_references` ?

aalexfvk (changes requested):

ch_backup/backup/deduplication.py
```diff
@@ -156,20 +134,22 @@ def handled(self):
...
-    dedup_info: DedupInfo,
-    layout: BackupLayout,
-    hostname: str,
+    context: BackupContext,
     dedup_backups_with_light_meta: List[BackupMetadata],
     databases: Sequence[Database],
 ) -> None:
     # pylint: disable=too-many-locals,too-many-branches
+    layout = context.backup_layout
+    dedup_info = _create_dedup_references()
```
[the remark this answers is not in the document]
kirillgarbar: This dict is used [here](https://github.com/yandex/ch-backup/blob/7ce04913e32d553ee8f2d9df3f9e4f7f2dbce1ef/ch_backup/backup/deduplication.py#L184) to check if processed part is already collected for deduplication, for example from previous metadata file. Not sure if it is possible if deduplication works correctly, but this check was there before.
We can use clickhouse table to check that by collecting part's metadata in batch and filtering them with JOIN query.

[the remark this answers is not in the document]
aalexfvk: Ok. But let's add a comment about it.

PR #170, opened 2024-07-08
Encrypt backup metadata

kirillgarbar:


ch_backup/backup/layout.py
```diff
@@ -298,11 +309,12 @@ def get_backup(
...
+        # New backup metadata is encrypted
+        # Retry in case it is old and not encrypted
         try:
-            data = self._storage_loader.download_data(path)
-            return BackupMetadata.load_json(data)
-        except Exception as e:
-            raise StorageError("Failed to download backup metadata") from e
+            return self._load_metadata(path, not use_light_meta)
+        except CryptoError:
+            return self._load_metadata(path, False)
```
aalexfvk, line 317: Now it seems it is better to do it with a flag for consistency with data encryption

aalexfvk: It is decided to keep the current approach, to avoid downloading light metadata every time, especially when all backups will be encrypted.
But let's add log message here to discover cases when file is actually corrupted.

aalexfvk (approved):

PR #171, opened 2024-07-12
Set fixed restore method for table engines and refactor

MikhailBurdukov:


ch_backup/clickhouse/control.py
```diff
@@ -575,19 +559,14 @@ def get_tables(
...
-        query_sql = GET_TABLE_SQL.format(
-            db_name=escape(db_name), table_name=escape(table_name)
-        )
-
-        result = self._ch_client.query(query_sql)["data"]
-        if result:
-            return self._make_table(result[0])
-
-        return None
+        tables = self.get_tables(db_name, [table_name], short_query)
```
aalexfvk, line 568: Let's add an exception if it is returned > 1

ch_backup/clickhouse/models.py
```diff
@@ -83,6 +83,64 @@ def _map_paths_to_disks(
...
+        Return True if table engine belongs to replicated merge tree table engine family, or False otherwise.
+        """
+        return self.is_merge_tree() and self.engine.find("Replicated") != -1
+
+    @staticmethod
+    def engine_is_replicated(engine: str) -> bool:
+        """
+        A static method for determining whether an engine is replicated or not.
+        """
+        return Table("", "", engine, [], [], "", "", None).is_replicated()
```
aalexfvk, line 97: Must be adjusted on every change of `Table` constructor.
Why not just `"MergeTree" in engine and "Replicated" in engine` and reuse it in `is_replicated()`

ch_backup/logic/table.py
```diff
@@ -654,20 +648,33 @@ def _restore_data(
...
+                assert (
+                    maybe_table_short is not None
+                ), f"Table not found {table_meta.database}.{table_meta.name}"
+
+                # We have to check table engine on short Table version
+                # because some of columns might be inaccessbible, for old ch versions.
+                # Fix https://github.com/ClickHouse/ClickHouse/pull/55540 is pesented since 23.8.
+                if not maybe_table_short.is_merge_tree():
+                    logging.debug(
+                        'Skiptable "{}.{}" data restore, because it is not MergeTree like.',
```
aalexfvk, line 663:
```suggestion
                        'Skip table "{}.{}" data restore, because it is not MergeTree family.',
```

ch_backup/logic/table.py
```diff
@@ -654,20 +648,33 @@ def _restore_data(
         for table_meta in tables:
             cloud_storage_parts = []
             try:
+                maybe_table_short = context.ch_ctl.get_table(
+                    table_meta.database, table_meta.name, short_query=True
+                )
+                assert (
```
aalexfvk, line 654: Let's use `if` and an exception instead of `assert`. `maybe_table_short` can be `None`, but asserts can be disabled in runtime

kirillgarbar (changes requested):

kirillgarbar (approved):

aalexfvk (approved):

PR #173, opened 2024-07-18
Check if backup name exists and add retry on ClickhouseError for queries

kirillgarbar:
Retries are added in case of OOM on SELECT and something similar, but we decided that retry can be set for every type of exception.

### DIFF
--- a/ch_backup/clickhouse/client.py
+++ b/ch_backup/clickhouse/client.py
@@ -37,7 +37,7 @@ class ClickhouseClient:
         """
         return self._session.params
 
-    @retry(requests.exceptions.ConnectionError)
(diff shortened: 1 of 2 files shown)


ch_backup/clickhouse/client.py
```diff
@@ -37,7 +37,7 @@ def settings(self):
         """
         return self._session.params
 
-    @retry(requests.exceptions.ConnectionError)
+    @retry((requests.exceptions.ConnectionError, ClickhouseError))
```
kirillgarbar, line 40: Retry of a failed non-SELECT queries will cause unnecessary logs and longer fail time, in that case a check if query is a SELECT query can be added with retries on `ClickhouseSelectError` or something similar

aalexfvk (approved):

PR #181, opened 2024-08-22
Add optional is_enabled flag for storage encryption

ivsnk:
Flag is_enabled is added for having ability to disable storage encryption, if it is not necessary.


tests/integration/features/partially_encrypted_backups.feature
```diff
@@ -0,0 +1,147 @@
...
+    When metadata paths of clickhouse01 backup #0 was deleted
+    """
+    - databases.test_db.engine
+    - databases.test_db.metadata_path
+    """
+    Given file "metadata/test_db/table_01.sql" in clickhouse01 backup #0 data set to
+    """
+    CREATE TABLE test_db.table_01 ON CLUSTER 'default' (date Date, n Int32)
+    ENGINE = ReplicatedMergeTree('/clickhouse/tables/test_db/table', '{replica}') PARTITION BY date ORDER BY date
+    """
```
aalexfvk, line 41: Why do we do this here for purposes of this test ?

aalexfvk:
How about testing the deduplication with previously encrypted backups ?

ivsnk:
> How about testing the deduplication with previously encrypted backups ?

Scenario "Restore backup after encryption disabled with valid metadata" performs encryption of the backup, change encryption flag -> false, and try to decrypt successfully with check that the flag in meta is still true. Is not it the same scenario as you are suggesting? Describe please your scenario, if it is not the same scenario

aalexfvk:
Not quite. Firstly, why do you use `noop` encryption instead of default `nacl` ? Better checking on encrypted data.
As for deduplication, I meant the case when we create two backup and the second (not encrypted) contains links to the parts from first backup (encrypted). 
Tests with deduplication can be found here https://github.com/yandex/ch-backup/blob/0c7fadc67ebfc22b82601be5d4f563375e8d04c5/tests/integration/features/deduplication.feature#L32

ch_backup/clickhouse/control.py
```diff
@@ -182,7 +182,8 @@
         files Array(String),
         tarball Bool,
         disk_name String,
-        verified Bool
+        verified Bool,
+        encrypted Nullable(Bool)
```
aalexfvk, line 186: Why `Nullable` and it seems [here](https://github.com/yandex/ch-backup/blob/849cd5b9135b0e5915dd7e21331a469b81d9a900/ch_backup/clickhouse/control.py#L1009) we should do DROP instead of TRUNCATE to be able to change the schema of this temporal table.

ivsnk: Nullable for backward compatibility (because previous tables do not contain this field)

ch_backup/clickhouse/control.py
```diff
@@ -757,6 +763,7 @@ def scan_frozen_parts(
                 checksum,
                 size,
                 rel_paths,
+                backup_meta.encrypted,
```
ivsnk, line 766: @aalexfvk As far as I understand here should be used encrypted field from certain data part, but it is not clear for me, how to get part metadata, could you please suggest some variants? (I have tried to get it from backup_meta using find_part and get_parts functions, but results of these functions were empty)

ch_backup/clickhouse/control.py
aalexfvk: Actually `_deduplication_info` table is temporal and recreated every run. I propose not to make this column just `Bool`

ch_backup/clickhouse/control.py
aalexfvk: I think we should not add `encrypted` field to `FrozenPart` at all, because frozen parts on disk are never encrypted. Where do you use

ch_backup/backup/metadata/part_metadata.py
```diff
@@ -134,4 +145,5 @@ def from_frozen_part(cls, frozen_part: FrozenPart) -> "PartMetadata":
             files=frozen_part.files,
             tarball=True,
             disk_name=frozen_part.disk_name,
+            encrypted=frozen_part.encrypted,
```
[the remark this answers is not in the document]
aalexfvk: @ivsnk Hello, you requested a review but what about this note ?

PR #185, opened 2024-09-10
MDB-30734 Metadata cleanup with system drop replica

MikhailBurdukov:


ch_backup/clickhouse/metadata_cleaner.py
```diff
@@ -0,0 +1,125 @@
...
+        """
+        if not self._ch_ctl.ch_version_ge("23.3"):
+            logging.warning(
+                "Ch version is too old, will skip replicated database cleanup."
+            )
+            return
+
+        replicated_databases_paths = get_database_zookeeper_paths(replicated_databases)
+
+        for [database, database_path, shard] in replicated_databases_paths:
```
aalexfvk, line 97:
```suggestion
        for database, database_path, shard in replicated_databases_paths:
```

ch_backup/clickhouse/metadata_cleaner.py
```diff
@@ -0,0 +1,125 @@
...
+        """
+        Remove replica tables metadata from zookeeper.
+        """
+        replicated_table_paths = get_table_zookeeper_paths(replicated_tables)
+
+        for table, table_path in replicated_table_paths:
+            table_macros = dict(
+                database=table.database, table=table.name, uuid=table.uuid
+            )
+            table_macros.update(self._macros)
```
aalexfvk, line 63: Does this imply that `self._macros` must not contain `database`, `table` and `uuid` keys ? What if they exists for some reason

ch_backup/logic/database.py
```diff
@@ -56,6 +60,14 @@ def restore(
                 databases_to_restore[name] = db
                 continue
 
+        if metadata_cleaner:
+            replicated_databases = [
+                database
+                for [_, database] in databases_to_restore.items()
```
aalexfvk, line 66:
```suggestion
                for database in databases_to_restore.values()
```

tests/integration/features/backup_replicated.feature
```diff
@@ -786,3 +786,72 @@ Feature: Backup replicated merge tree table
...
+      | 0   | created | 10         | 0          |
+
+    When we stop clickhouse at clickhouse01
+    When we restore clickhouse backup #0 to clickhouse02
+    """
+    replica_name: clickhouse01
+    schema_only: true
+    """
+    When we start clickhouse at clickhouse01
+    Then replica test_db.table_01 on clickhouse01 is read-only
```
aalexfvk, line 821: Can we do more specific check of metadata cleaning, not just table is in RO ?

ch_backup/clickhouse/metadata_cleaner.py
MikhailBurdukov: Added more strict order of macroses update.

But if such macroses in `self._macros` it is a clickhouse misconfiguration, because they are exactly the same as macroses in the ch.

aalexfvk (approved):

PR #189, opened 2024-09-30
Support for ch-24.9

MikhailBurdukov:


MikhailBurdukov:
@Alex-Burmak tests seems fine. Can we merge?

PR #191, opened 2024-10-04
Bugfix: order parts after backup for ReplacingMergeTree 

k-morozov:
Test that reproduce problem was added.

```
    When we execute query on clickhouse02                         # tests/integration/steps/clickhouse.py:106 0.006s
      """
      SELECT id, visits FROM test_db.hits FINAL ORDER BY id FORMAT Vertical;
      """
    Then we get response                                          # tests/integration/steps/clickhouse.py:132 0.000s
      """
      Row 1:
      ──────
      id:     1
      visits: 106
      """
      Assertion Failed: 
      Expected: 'Row 1:\n──────\nid:     1\nvisits: 106'
           but: was 'Row 1:\n──────\nid:     1\nvisits: 105'
```

### DIFF
--- a/tests/integration/features/backup_restore.feature
+++ b/tests/integration/features/backup_restore.feature
@@ -87,6 +87,78 @@ Feature: Backup & Restore
     And we restore clickhouse backup #0 to clickhouse02
     Then we got same clickhouse data at clickhouse01 clickhouse02
 
+  Scenario: Backup & Restore for ReplacingMergeTree
+    When we drop all databases at clickhouse01
+    And we drop all databases at clickhouse02
+    Given we have executed queries on clickhouse01
+    """
+    CREATE DATABASE test_db;
+    CREATE TABLE test_db.hits (
+        id UInt32,
+        url String,
+        visits UInt32
+    )
+    ENGINE ReplacingMergeTree
+    ORDER BY id;
+    INSERT INTO test_db.hits VALUES
+      (1, '/index', 100);
+    """
+    Given we have executed queries on clickhouse01
+    """
+    INSERT INTO test_db.hits VALUES
+      (1, '/index', 101);
+    """
+    Given we have executed queries on clickhouse01
+    """
+    INSERT INTO test_db.hits VALUES
+      (1, '/index', 102);
+    """
+    Given we have executed queries on clickhouse01
+    """
+    INSERT INTO test_db.hits VALUES
+      (1, '/index', 103);
+    """
+    Given we have executed queries on clickhouse01
+    """
+    INSERT INTO test_db.hits VALUES
+      (1, '/index', 104);
+    """
+    Given we have executed queries on clickhouse01
+    """
+    INSERT INTO test_db.hits VALUES
+      (1, '/index', 105);
+    """
+    Given we have executed queries on clickhouse01
+    """
+    INSERT INTO test_db.hits VALUES
+      (1, '/index', 106);
+    """
+    When we create clickhouse01 clickhouse backup
+    And we restore clickhouse backup #0 to clickhouse02
+    When we execute query on clickhouse01
+    """
+    SELECT id, visits FROM test_db.hits FINAL ORDER BY id FORMAT Vertical;
+    """
+    Then we get response
+    """
+    Row 1:
+    ──────
+    id:     1
+    visits: 106
+    """
+    When we execute query on clickhouse02
+    """
+    SELECT id, visits FROM test_db.hits FINAL ORDER BY id FORMAT Vertical;
+    """
+    Then we get response
+    """
+    Row 1:
+    ──────
+    id:     1
+    visits: 106
+    """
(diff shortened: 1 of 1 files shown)


ch_backup/backup/metadata/table_metadata.py
```diff
@@ -53,6 +55,24 @@ def get_parts(self, *, excluded_parts: Set[str] = None) -> List[PartMetadata]:
                     PartMetadata.load(self.database, self.name, part_name, raw_metadata)
                 )
 
+        if self.engine == REPLACING_MERGE_TREE:
```
Alex-Burmak, line 58: What about ReplicatedReplacingMergeTree and table engines like CollapsingMergeTree? Wouldn't it better to perform the code below unconditionally?

ch_backup/backup/metadata/table_metadata.py
```diff
@@ -53,6 +55,24 @@ def get_parts(self, *, excluded_parts: Set[str] = None) -> List[PartMetadata]:
...
+                suffix = ""
+                try:
+                    partition = chunks[0]
+                    start_range = int(chunks[1])
+                    finish_range = int(chunks[2])
+                    suffix = chunks[3]
+                except (IndexError, ValueError):
+                    start_range = 0
+                    finish_range = 0
+                return partition, start_range, finish_range, suffix
```
aalexfvk, line 72: Let's use `from typing import NamedTuple` for clarity. It is also lexicographically comparable as a usual tuple

ch_backup/backup/metadata/table_metadata.py
```diff
@@ -53,6 +55,24 @@ def get_parts(self, *, excluded_parts: Set[str] = None) -> List[PartMetadata]:
                     PartMetadata.load(self.database, self.name, part_name, raw_metadata)
                 )
 
+        if self.engine == REPLACING_MERGE_TREE:
+
+            def split_part_name(part):
+                chunks = part.split("_", maxsplit=3)
```
aalexfvk, line 61: Let's parse full format `{partition_id}_{min_block}_{max_block}_{level}_{mutation}`
In accordance to https://github.com/ClickHouse/ClickHouse/blob/e2821c5e8b728d1d28f9e0b98db87e0af5bc4a29/src/Storages/MergeTree/MergeTreePartInfo.cpp#L54

ch_backup/backup/metadata/table_metadata.py
```diff
@@ -53,6 +55,24 @@ def get_parts(self, *, excluded_parts: Set[str] = None) -> List[PartMetadata]:
...
+                    partition = chunks[0]
+                    start_range = int(chunks[1])
+                    finish_range = int(chunks[2])
+                    suffix = chunks[3]
+                except (IndexError, ValueError):
+                    start_range = 0
+                    finish_range = 0
+                return partition, start_range, finish_range, suffix
+
+            result.sort(key=lambda part: split_part_name(part.name))
```
aalexfvk, line 74: Please, update the docstring of the function that it returns list in the sorted order. 
BTW, wherever this function is called, do we need a sorted order?
Maybe enable that by parameter or make as a separate function.

ch_backup/backup/metadata/table_metadata.py
MikhailBurdukov: And pls  add this link in the comments.

ch_backup/backup/metadata/table_metadata.py
k-morozov: Good idea.

ch_backup/backup/metadata/table_metadata.py
k-morozov: Ok

ch_backup/backup/metadata/table_metadata.py
k-morozov: ok

ch_backup/backup/metadata/table_metadata.py
k-morozov: I think we dont need get_parts without sorting. I suggest using sorted parts as result this function in all cases.

ch_backup/backup/metadata/table_metadata.py
```diff
@@ -53,6 +66,25 @@ def get_parts(self, *, excluded_parts: Set[str] = None) -> List[PartMetadata]:
                     PartMetadata.load(self.database, self.name, part_name, raw_metadata)
                 )
 
+        def split_part_name(part: str) -> PartInfo:
+            max_split = 4
+            chunks = part.split("_", maxsplit=max_split)
+            partition_id = ""
+            level = 0
+            mutation = ""
```
aalexfvk, line 74: mutation version is also the number

k-morozov: ok

aalexfvk (approved):

PR #213, opened 2025-03-18
named collections ddl from zookeeper

goncharovnikita:


tests/integration/features/named_collections.feature
```diff
@@ -11,16 +11,27 @@ Feature: Backup and restore functionality of named collections
...
+    Then we got the following backups on clickhouse01
+      | num | state   | data_count | link_count |
+      | 0   | created | 0          | 0          |
+    When we restore clickhouse backup #0 to clickhouse02
+    Given a working clickhouse on clickhouse02
+    Then clickhouse01 has same named collections as clickhouse02
+    And clickhouse01 has same schema as clickhouse02
+
+  @require_version_24.9
   Scenario: from local to local storage
```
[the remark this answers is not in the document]
goncharovnikita: Fixed, made them table-driven with hot config-swapping

ch_backup/clickhouse/models.py
```diff
@@ -275,3 +279,58 @@ def __init__(
...
+    def iv_hex(self):
+        """
+        Encrypted file header initial vector.
+        """
+        return self._iv.hex()
+
+
+class EncryptedFile(SimpleNamespace):
+    """
+    Encrypted file.
```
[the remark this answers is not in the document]
goncharovnikita: NamedCollectionsStorage restricted to AES-CTR encryption at the moment, however EncryptedFile represents just encrypted file, as ClickHouse stores it, it does not strictly related to NamedCollectionsStorage

So i suggest to leave this generalized name

ch_backup/encryption/util.py
```diff
@@ -0,0 +1,47 @@
+"""
```
goncharovnikita, line 1: Talked with @MikhailBurdukov offline, decided to move this file into ch_backup/clickhouse/encryption.py

tests/integration/features/named_collections.feature
```diff
@@ -1,27 +1,27 @@
 Feature: Backup and restore functionality of named collections
 
-  Background:
+  @require_version_24.9
+  Scenario Outline: create and restore named collections
```
MikhailBurdukov, line 4:
```
Scenario Outline: create and restore named collections: {name}
```
Where `name` should be column in Examples

ch_backup/zookeeper/util.py
```diff
@@ -0,0 +1,34 @@
+"""
+Zookeeper related utilities.
+"""
+
+import os
+
+from ch_backup.zookeeper.zookeeper import ZookeeperCTL
+
+
+def copy_directory_content_from_zookeeper(
```
MikhailBurdukov, line 10: actually i don't really like this method.
Because it will work only in our's scenario, not in general case. Because node in zk != directory.
 I would prefer just `for in zk_client.get_chilren`.

But I'll leave it to you, if you want to keep it, lets move it to zookeeper.py

ch_backup/clickhouse/config.py
```diff
@@ -35,3 +37,79 @@ def config(self) -> dict:
...
+        """
+        Create NamedCollectionsStorageConfig from ClickhouseConfig.
+        """
+        nc_config = ch_config.config.get("named_collections_storage")
+        if not nc_config:
+            storage_path = ch_backup_config.get("named_collections_path")
+
+            assert storage_path, "named_collections_path missing from ch-backup config"
+
+            return NamedCollectionsStorageConfig(storage_path=storage_path)
```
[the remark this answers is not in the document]
goncharovnikita: It's for cases, when ClickHouse configuration miss the `named_collections_storage`

MikhailBurdukov (approved):
lgmt

ch_backup/logic/named_collections.py
```diff
@@ -83,3 +139,108 @@ def get_named_collections_list(context: BackupContext) -> List[str]:
...
+        self,
+        zk_ctl: ZookeeperCTL,
+        from_path_dir: str,
+        to_path_dir: str,
+    ) -> None:
+        """
+        Copy all files from zookeeper directory to destination.
+        """
+        if to_path_dir[-1] != "/":
+            to_path_dir += "/"
```
MikhailBurdukov, line 153: Looks unnecessary. 
Haven't  found in [docs](https://docs.python.org/3/library/os.path.html), that path must be ended with '/'
```
>>> os.path.join('/a/b/c/', 'd')
'/a/b/c/d'
>>> os.path.join('/a/b/c', 'd')
'/a/b/c/d'
```

aalexfvk (approved):

PR #214, opened 2025-03-21
Don't select data_paths from StorageProxy

kirillgarbar:


aalexfvk (approved):

kirillgarbar:
In CH version < 24.3 adding `WHERE engine != 'StorageProxy'` still throws exception and only filtering by name works. Had to add the query for older versions.

PR #219, opened 2025-04-08
Freeze table by partitions

kirillgarbar:


ch_backup/clickhouse/control.py
```diff
@@ -548,6 +565,51 @@ def freeze_table(self, backup_name: str, table: Table) -> None:
...
+        Make snapshot of the specified table.
+        """
+        # Table has no partitions or created with deprecated syntax.
+        # FREEZE PARTITION ID with deprecated syntax throws segmentation fault in CH.
+        if "PARTITION BY" not in table.create_statement:
+            return self.freeze_table(backup_name, table)
+
+        partitions_to_freeze = self.list_partitions(table)
+
+        with ThreadPoolExecutor(max_workers=threads) as pool:
```
aalexfvk, line 581: How about reusing [ExecPool](https://github.com/yandex/ch-backup/blob/df63bd223241d4e65c24fc09e97e1038d4fa2775/ch_backup/storage/async_pipeline/base_pipeline/exec_pool.py#L24) facility. It will simplify a little and add some logging.

ch_backup/clickhouse/control.py
```diff
@@ -548,6 +565,51 @@ def freeze_table(self, backup_name: str, table: Table) -> None:
...
+            for partition in partitions_to_freeze:
+                query_sql = FREEZE_PARTITION_SQL.format(
+                    db_name=escape(table.database),
+                    table_name=escape(table.name),
+                    backup_name=backup_name,
+                    partition=partition,
+                )
+                futures.append(
+                    pool.submit(
+                        self._ch_client.query,
```
aalexfvk, line 592: Please validate thread-safety of `self._ch_client`

ch_backup/logic/table.py
```diff
@@ -133,33 +135,33 @@ def _backup(
...
-            with ThreadPoolExecutor(max_workers=freeze_threads) as pool:
+            with ThreadExecPool(
+                multiprocessing_config.get("freeze_threads", 1)
+            ) as pool:
                 for table in tables_:
-                    future = pool.submit(
+                    pool.submit(
+                        f'Freeze table "{table.database}"."{table.name}"',
                         TableBackup._freeze_table,
+                        None,
```
aalexfvk, line 145: Let's make it keyword-only argument to avoid passing opaque `None`

ch_backup/logic/table.py
```diff
@@ -174,13 +176,15 @@ def _freeze_table(
...
-    ) -> Optional[Tuple[Table, bytes]]:
+        freeze_partition_threads: int,
+    ) -> Optional[Table]:
         """
         Freeze table and return it's create statement
         """
         logging.debug('Trying to freeze "{}"."{}"', table.database, table.name)
 
         create_statement = TableBackup._load_create_statement_from_disk(table)
+        table.create_statement = create_statement if create_statement else ""
```
aalexfvk, line 187:
```suggestion
        table.create_statement = create_statement or ""
```

ch_backup/clickhouse/client.py
```diff
@@ -68,13 +74,23 @@ def query(
...
                 timeout = self.timeout
 
-            response = self._session.post(
-                self._url,
-                params=settings,
-                json=post_data,
-                timeout=(self.connect_timeout, timeout),
-                data=query.encode("utf-8"),
-            )
+            if new_session:
```
aalexfvk, line 77: Let's make something like this (not tested):
```
@contextmanager
def get_session(new_session: Optional[bool] = False):
    session = self._create_session(config, settings) if new_session else self._session
    try:
        yield session
    finally:
      if new_session:
        session.close()
```

ch_backup/clickhouse/client.py
```diff
@@ -110,3 +126,57 @@ def _create_session(config, settings):
         requests.packages.urllib3.disable_warnings()  # type: ignore[attr-defined]
 
         return session
+
+
+class ClickhouseClientMultithreading:
```
aalexfvk, line 131: Can we get rid of this class and pass client methods to pool directly there it is used.

ch_backup/clickhouse/control.py
aalexfvk: Write please summary of your investigation

ch_backup/clickhouse/client.py
kirillgarbar: The idea of this class was to enforce the need to use `new_session` where the client is working with multi-threading. If we ok with explicitly adding `new_session = True` when working with multi-threading, then this class can be removed.

aalexfvk: Yes it's ok with `new_session = True`

aalexfvk (approved):

PR #220, opened 2025-04-16
Set PartState.DOWNLOADED immediately for each part

kirillgarbar:


MikhailBurdukov (approved):
Generally LGTM except one renaming, lets wait for green CI:)

PR #223, opened 2025-04-28
Added ` NOT_DOWLOADED` state for parts.

MikhailBurdukov:
We have json with state of each part during restoring the backup and is's quite useful for big backups, helps to determine the current state of the restore(how many parts already restored, how many invalid parts and etc).  
But right now default value for the part state are `INVALID`.  So it is impossible to understand is part really `invalid` or it hasn't downloaded yet.


aalexfvk (approved):

MikhailBurdukov:
> Please write a short description about motivation for the PR

Edited the PR description

PR #225, opened 2025-05-01
Update supported Python and ClickHouse versions

Alex-Burmak:


.github/workflows/workflow.yml
```diff
@@ -43,10 +43,6 @@ jobs:
       fail-fast: false
       matrix:
         include:
-        - python: "3.7"
-          clickhouse: "latest"
-        - python: "3.8"
-          clickhouse: "latest"
```
Alex-Burmak, line 49: Unsupported versions
https://devguide.python.org/versions/

requirements.txt
```diff
@@ -12,9 +12,7 @@ tenacity<8.4.0
 kazoo
 xmltodict
 pypeln==0.4.9
-dataclasses>=0.7,<0.8; python_version <"3.7"  # required for pypeln==0.4.9
-typing_extensions>=3.7.4,<4.0; python_version <"3.8"  # required for pypeln==0.4.9
 loguru
-setuptools>=71.1; python_version >="3.8"  # instead of python3-setuptools package
```
Alex-Burmak, line 18: Simplify specification of requirements as python 3.7 and python 3.8 are not supported any more.

PR #236, opened 2025-05-18
AI based code review

Alex-Burmak:


Alex-Burmak:
> Ensure the GitHub CLI is installed before using gh. Add a setup step (e.g.
cli/cli-action@v2) so the workflow won’t fail due to missing CLI.

GitHub CLI is preinstalled on all GitHub-hosted runners.

https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/using-github-cli-in-workflows

----

> Expand the trigger types to include reopened and synchronize so the workflow runs on
updates and reopens, not just when a PR is first opened.

The trigger types includes only `opened` to avoid potential spam in pr comments. If re-review is required after updating patch set, it can be easily requested manually through the command to `@CodiumAI-Agent`.

Alex-Burmak:
Closed in favour of other AI code review tool - Sourcery.

PR #239, opened 2025-05-21
Replace bandit linter with ruff linter security checks

Alex-Burmak:
## Summary by Sourcery

Replace Bandit with Ruff for security linting by removing Bandit config and targets, enabling Ruff security rules, and updating scripts, docs, and CI accordingly.

Enhancements:
- Consolidate security linting under Ruff by enabling security rule category “S” and adding specific checks (S101, S108, S110, S113, S311, S324, S602)
- Remove Bandit from development dependencies, Makefile targets, and project configuration
- Add per-file ignores to disable Ruff security checks in test files

CI:
- Remove Bandit lint step from GitHub Actions workflow

Documentation:
- Update README and Makefile help text to remove Bandit references and rename `start-env` to `start-test-env`
- Bump default ClickHouse version in README

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -29,8 +29,6 @@ jobs:
       run: make pylint
     - name: "lint: mypy"
       run: make mypy
-    - name: "lint: bandit"
-      run: make bandit
 
   test:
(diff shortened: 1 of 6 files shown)


Alex-Burmak:
Additional motivation for these changes is to unify linter set with https://github.com/yandex/ch-tools where we don't use bandit tool.

PR #242, opened 2025-05-24
Update dependencies: pynacl, mypy, deepdiff

Alex-Burmak:


ch_backup/clickhouse/control.py
```diff
@@ -939,7 +939,7 @@ def scan_frozen_parts(
 
         if not os.path.exists(path):
             logging.debug("Shadow path {} is empty", path)
-            return []
+            return
```
Alex-Burmak, line 942: It's fix for the mypy error:
```
ch_backup/clickhouse/control.py:942: error: No return value expected  [return-value]
```

This function is a generator (has `yield`). So `return` should be without a value.

ch_backup/clickhouse/models.py
```diff
@@ -149,7 +149,7 @@ def is_external_engine(self) -> bool:
             "URL",
         )
 
-    def __hash__(self):
+    def __hash__(self) -> int:  # type: ignore
         return hash((self.database, self.name))
```
Alex-Burmak, line 153: Suppression of the error from new mypy version.

```
ch_backup/clickhouse/models.py:152: error: Signature of "__hash__" incompatible with supertype "SimpleNamespace"  [override]
ch_backup/clickhouse/models.py:152: note:      Superclass:
ch_backup/clickhouse/models.py:152: note:          None
ch_backup/clickhouse/models.py:152: note:      Subclass:
ch_backup/clickhouse/models.py:152: note:          def __hash__(self) -> Any
```

PR #260, opened 2025-09-03
Rename tables with long names before drop

MikhailBurdukov:


ch_backup/logic/table.py
```diff
@@ -604,10 +608,39 @@ def _preprocess_tables_to_restore(
...
+                    ### If table name longer than `getMaxTableNameLengthForDatabase()` ch function result
+                    ### then we can't drop the table. Just rename it to some random name and drop the table.
+                    ### But for dictionaries doesn't works create/attach with long name, so we can't even restore the dictionary
+                    ### with long name. So it better to keep it.
                     if existing_table.is_dictionary():
                         context.ch_ctl.drop_dictionary_if_exists(existing_table)
                     else:
-                        context.ch_ctl.drop_table_if_exists(existing_table)
+                        ### The lightweight copy to that we can modify and use for ch queries
+                        table_to_drop = Table(
```
aalexfvk, line 619: I'd propose to add classmetod `Table.make_dummy(db, name)`, to hide all this empty parameters

ch_backup/logic/table.py
```diff
@@ -604,10 +608,39 @@ def _preprocess_tables_to_restore(
...
+                            [],
+                            "",
+                            "",
+                            None,
+                        )
+                        if (
+                            len(table.name)
+                            > context.config_root["restore"]["max_table_name"]
+                        ):
+                            new_table_name = "".join(
```
aalexfvk, line 633: Let's add some recognizable prefix to quickly recognize such tables during debug if it is not deleted

PR #314, opened 2026-04-23
[fix] Support default configuration

otselnik:
## Summary by Sourcery

Handle ClickHouse configurations without explicit storage_configuration disks and ensure missing cloud_storage disks produce clear errors.

Bug Fixes:
- Avoid KeyError in ClickHouseTemporaryDisks.__enter__ when storage_configuration or disks are absent in the ClickHouse config.
- Raise a descriptive ClickHouseDisksException when a disk referenced in backup cloud storage metadata is missing from the ClickHouse storage_configuration instead of failing with a bare KeyError.

Tests:
- Add unit tests covering ClickHouseTemporaryDisks behavior when storage_configuration is absent and when referenced disks are missing from the configuration.
- Add integration scenarios verifying successful restore on nodes without storage_configuration and clear error messaging when cloud_storage disks are missing from the target config.

Chores:
- Add ClickHouse config fixtures without storage_configuration and without the s3 disk for use in tests.

### DIFF
--- a/ch_backup/clickhouse/disks.py
+++ b/ch_backup/clickhouse/disks.py
@@ -74,7 +74,9 @@ class ClickHouseTemporaryDisks:
         self._ch_availible_disks: Dict[str, Disk] = {}
 
     def __enter__(self):
-        self._disks = self._ch_config.config.get("storage_configuration", {}).get("disks", {})
+        self._disks = self._ch_config.config.get("storage_configuration", {}).get(
+            "disks", {}
+        )
         for disk_name in self._backup_meta.cloud_storage.disks:
             self._create_temporary_disk(
                 self._backup_meta,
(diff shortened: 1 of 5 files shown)


ch_backup/clickhouse/disks.py
```diff
@@ -74,7 +74,9 @@ def __init__(
         self._ch_availible_disks: Dict[str, Disk] = {}
 
     def __enter__(self):
-        self._disks = self._ch_config.config["storage_configuration"]["disks"]
+        self._disks = (self._ch_config.config.get("storage_configuration") or {}).get(
+            "disks"
+        ) or {}
```
aalexfvk:
```suggestion
        self._disks = self._ch_config.config.get("storage_configuration", {}).get("disks", {})
```

aalexfvk (approved):

PR #321, opened 2026-05-14
MDB-44986 respect keep going while restore

UberDever:
## Summary by Sourcery

Improve table restore robustness and reporting when using the keep-going option during backup restore.

Bug Fixes:
- Prevent restore from aborting on missing required tables, missing create statements, duplicate UUIDs, or preprocessing errors when keep-going is enabled.
- Ensure detached tables are matched by UUID first and then by name to attach the correct table during restore.

Enhancements:
- Return structured preprocessing results for restore, including counts of skipped tables and the final table set to restore.
- Add detection and resolution of duplicate table UUIDs in backup metadata, with conflict handling and detailed logging.
- Improve restore preprocessing logging by summarizing skipped tables and reasons at the end of preprocessing.

Tests:
- Extend unit tests to cover restore preprocessing behavior, including duplicate UUID handling, detached table matching, and accumulation of skipped tables under keep-going.
- Add an integration scenario validating restore behavior with conflicting table UUIDs and the effect of the keep-going flag.


ch_backup/logic/table.py
```diff
@@ -358,39 +365,59 @@ def restore(
...
+                    "Required tables {} were not found in backup metadata, skipping due to --keep-going flag",
+                    missed_tables_names,
                 )
-
-            logging.debug("All required tables are present")
+            else:
+                logging.debug("All required tables are present")
 
             logging.debug("Leaving only required tables metadata")
             required_tables = [(table.database, table.name) for table in tables]
```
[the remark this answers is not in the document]
UberDever: ur lame, its cool

ch_backup/logic/table.py
```diff
@@ -652,84 +658,118 @@ def _preprocess_tables_to_restore(
...
+                    and (table.database, table.name) not in readonly_cleaned_metadata
+                ):
+                    logging.debug(
+                        f"Will clean ZooKeeper metadata for table {table_name_for_logs}"
+                    )
+                    tables_to_clean_metadata.append(table)
+            except (ClickhouseBackupError, ClickhouseError, OSError, ValueError) as e:
+                if not keep_going:
+                    raise
+                preprocessing_failed_skipped += 1
```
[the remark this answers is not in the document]
UberDever: We expect only these 4 errors that can occur as a result of other-party activity (disk, clickhouse inner workings, ...), so other errors are considered a bug in the ch-backup, in which case we should fail

ch_backup/logic/table.py
```diff
@@ -652,84 +662,118 @@ def _preprocess_tables_to_restore(
...
+                    logging.warning(
+                        'Table {} will be recreated as its schema mismatches the schema from backup: "{}" != "{}"',
+                        table_name_for_logs,
+                        existing_table.create_statement,
+                        table.create_statement,
+                    )
+                elif existing_table := (
+                    existing_tables_by_uuid.get(table.uuid) if table.uuid else None
+                ):
+                    logging.warning(
```
aalexfvk, line 718: Let's restore this comment 
`# Schemas mismatch here at least in name`

ch_backup/logic/table.py
```diff
@@ -863,11 +907,103 @@ def _get_tables_from_meta(
...
+                "Create statement for table {}.{} is missing in backup layout, skipping due to --keep-going flag",
+                table.database,
+                table.name,
+            )
 
-        return list(tables.values())
+        return result, missing_create_statement_skipped
+
+    @staticmethod
+    def _resolve_duplicate_backup_tables(
```
aalexfvk: It is out of scope but useful. Can we do this inside table loop inside `_preprocess_tables_to_restore` ?

tests/unit/test_backup_tables.py
```diff
@@ -229,3 +229,244 @@ def test_validate_all_parts_checked_before_raising(self):
             TableBackup._validate_uploaded_parts(context, parts)
 
         assert check_mock.call_count == 3
+
```
aalexfvk, line 232: Let's remove unit tests — tests with a large number of mocks and patches for `_functions` are evil.

UberDever: Well, I think the same way, but on the other hand we have very clear contract for restore function here: what will be skipped and why. I agree that this increases coupling tho
We can remove them if contracts are not so important in this case in particular

ch_backup/logic/table.py
UberDever: I think it can be inlined, yes

aalexfvk: I meant inline to `_preprocess_tables_to_restore`

tests/unit/test_backup_tables.py
aalexfvk: If there is some edge cases lets cover them by BEHAVE tests

aalexfvk (approved):

PR #332, opened 2026-06-29
Detect concurrent object overwrite by comparing ETag for download

aalexfvk:
## Summary by Sourcery

Detect and handle concurrent overwrites of S3 objects during multipart download by tracking ETags and enforcing consistent streams.

Bug Fixes:
- Prevent returning a mixed data stream when an S3 object is overwritten mid-download by comparing part ETags and raising an error.
- Avoid off-by-one and boundary issues in multipart download range handling and ensure end-of-stream is detected correctly.

Enhancements:
- Use head_object to initialize multipart downloads with object metadata, including ETag, and restart downloads safely if the object changes before reading begins.
- Return None when a downloaded part is empty to signal completion consistently with other end-of-download conditions.

### DIFF
--- a/ch_backup/storage/engine/s3/s3_engine.py
+++ b/ch_backup/storage/engine/s3/s3_engine.py
@@ -10,6 +10,8 @@ from typing import Optional, Sequence
 import requests
 from botocore.exceptions import ClientError
 
+from ch_backup import logging
+from ch_backup.exceptions import StorageError
 from ch_backup.storage.engine.base import PipeLineCompatibleStorageEngine
 from ch_backup.storage.engine.s3.s3_client_factory import (
     S3ClientCachedFactory,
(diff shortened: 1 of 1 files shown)


ch_backup/storage/engine/s3/s3_engine.py
```diff
@@ -195,34 +197,65 @@ def complete_multipart_upload(self, remote_path: str, upload_id: str) -> None:
...
+
+        # Detect concurrent object overwrite by comparing ETag.
+        part_etag = part.get("ETag")
+        if part_etag and part_etag != download["etag"]:
+            if range_start == 0:
+                # No data delivered yet — the object was overwritten before we
+                # started reading.  Update metadata and continue with the new
+                # version so the caller gets a consistent response.
+                logging.warning(
+                    "Object '{}' was overwritten before download started "
```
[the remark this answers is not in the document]
aalexfvk: Custom logger is used in the project. It's ok

ch_backup/storage/engine/s3/s3_engine.py
```diff
@@ -195,34 +197,65 @@ def complete_multipart_upload(self, remote_path: str, upload_id: str) -> None:
...
+                )
+                # ContentRange format: "bytes <start>-<end>/<total>"
+                content_range = part.get("ContentRange", "")
+                if content_range:
+                    download["total_size"] = int(content_range.rsplit("/", 1)[1])
+                download["etag"] = part_etag
+            else:
+                # Already delivered chunks of the old version — the data stream
+                # is now inconsistent.
+                raise StorageError(
```
MedvedewEM, line 250: Should we retry one more time to download file?

aalexfvk: Sure, but at what level would you suggest doing this to avoid overcomplicating the PR?

It looks like for small files (metadata), which are downloaded in one go, we've fixed the issue by updating the etag, while large files shouldn't be overwritten, and now we'll at least be able to catch such a situation and investigate it.

MedvedewEM (approved):

PR #333, opened 2026-07-01
Fix UDF restore idempotency

NikitaUnisikhin:
## Summary by Sourcery

Tests:
- Add integration test to verify UDF restore remains idempotent when applied repeatedly from the same backup on ClickHouse 21.11.

### DIFF
--- a/ch_backup/clickhouse/control.py
+++ b/ch_backup/clickhouse/control.py
@@ -952,9 +952,7 @@ class ClickhouseCTL:
         """
         Restore user defined function.
         """
-        self._ch_client.query(
-            udf_statement.replace("CREATE FUNCTION", "CREATE OR REPLACE FUNCTION", 1)
-        )
+        self._ch_client.query(udf_statement)
 
     def restore_named_collection(self, nc_statement):
         """
--- a/ch_backup/clickhouse/control.py
+++ b/ch_backup/clickhouse/control.py
@@ -952,13 +952,19 @@ class ClickhouseCTL:
         """
         Restore user defined function.
         """
-        self._ch_client.query(udf_statement)
+        self._ch_client.query(
+            udf_statement.replace("CREATE FUNCTION", "CREATE OR REPLACE FUNCTION", 1)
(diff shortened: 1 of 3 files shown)


ch_backup/clickhouse/control.py
```diff
@@ -952,13 +952,19 @@ def restore_udf(self, udf_statement):
         """
         Restore user defined function.
         """
-        self._ch_client.query(udf_statement)
+        self._ch_client.query(
```
NikitaUnisikhin: https://github.com/yandex/ch-backup/blob/51b0c007761d56d50b8caa5bf526b49171adb043/ch_backup/clickhouse/client.py#L55 - query has a retry decorator. If a network disruption occurs after the UDF is created, the request will be repeated and an error "object '***' already exists" will be displayed.

ch_backup/clickhouse/control.py
```diff
@@ -952,7 +952,9 @@ def restore_udf(self, udf_statement):
...
+            udf_statement.replace("CREATE FUNCTION", "CREATE OR REPLACE FUNCTION", 1)
```
aalexfvk: Let's move this logic to `UDFBackup` class as `_rewrite_udf_schema` function by analogy with `TableBackup._rewrite_table_schema`

aalexfvk (changes requested):

ch_backup/logic/udf.py
```diff
@@ -43,25 +43,13 @@ def restore(self, context: BackupContext) -> None:
...
-        udf_on_clickhouse_list = udf_on_clickhouse.keys()
-
         for udf_name in udf_list:
             logging.debug("Restoring UDF {}", udf_name)
 
             statement = context.backup_layout.get_udf_create_statement(
                 context.backup_meta, udf_name
             )
-
-            if (
```
aalexfvk, line 56: It looks like we can retain this logic.

aalexfvk (approved):

PR #351, opened 2026-09-02
MDB-48452: Fix race on named collections restore

kirillgarbar:
## Summary by Sourcery

Make named collection restoration race-safe and idempotent by preserving identical collections and safely recreating only changed definitions.

Bug Fixes:
- Prevent named collection restore races by comparing the existing persisted DDL and avoiding unnecessary drop/recreate operations when the collection is unchanged.

Enhancements:
- Make named collection restoration idempotent across local, ZooKeeper, and encrypted storage configurations.
- Normalize CREATE statements and use IF NOT EXISTS for supported ClickHouse versions.

Tests:
- Add unit and integration coverage for idempotent restoration and identical named collections in encrypted ZooKeeper storage.

### DIFF
--- a/ch_backup/logic/named_collections.py
+++ b/ch_backup/logic/named_collections.py
@@ -4,14 +4,18 @@ Clickhouse backup logic for named collections
 
 import os
 import posixpath
+import re
 from dataclasses import dataclass, field
 from enum import Enum
-from typing import List
+from pathlib import Path
+from typing import List, Optional
+
+from kazoo.exceptions import NoNodeError
 
 from ch_backup import logging
 from ch_backup.backup_context import BackupContext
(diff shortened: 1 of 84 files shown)


ch_backup/logic/named_collections.py
```diff
@@ -140,6 +159,95 @@ def get_named_collections_list(context: BackupContext) -> List[str]:
         """
         return context.backup_meta.get_named_collections()
 
+    @staticmethod
+    def _add_if_not_exists(statement: str) -> str:
```
aalexfvk, line 166: `IF NOT EXISTS` is only supported for >= 23.8. Let's check our target versions.

aalexfvk (approved):

PR #365, opened 2026-09-15
Retry checksum read on fs/disk errors

UberDever:
## Summary by Sourcery

Improve frozen-part checksum handling by retrying transient disk errors and reporting unrecoverable read failures with backup context.

Bug Fixes:
- Retry checksum reads up to three times when transient filesystem I/O errors occur.
- Report checksum-read failures as contextual Clickhouse backup errors so failed backups expose the affected part and table.

Tests:
- Add coverage for retry behavior across open and read failures and for contextual checksum error propagation.


tests/unit/test_control.py
```diff
@@ -1,6 +1,18 @@
+import errno
```
aalexfvk, line 1: Let’s remove these unit tests: they mock private functions and are too implementation-specific.

UberDever: yes, the only thing I would keep is the fs mock to not actually interact with an FS at all

aalexfvk:
Closing this PR after confirming persistent read errors on /dev/vda. Retrying EIO in ch-backup would only delay detection of a degraded disk. The correct action is to replace or evacuate the disk, or run the backup from a healthy replica.

PR #370, opened 2026-09-21
MDB-49626: encryption and compression options do not work

roaldm153:
In `table.py`:

```python
if context.cloud_conf.get("cloud_storage", {}).get("encryption", True):
    ...
if context.cloud_conf.get("cloud_storage", {}).get("compression", True):
    ...
```

And in `backup_context.py`:

```python
self._cloud_conf = config.get("cloud_storage")
```

The issue is that there is no `cloud_storage` key in `_cloud_conf`. Thus, `_.get("cloud_storage", {})` always returns an empty dictionary, and `_.get("...", True)` always returns the default value (i.e., `True`).

As a result, the encryption and compression options are effectively disabled. Cloud storage metadata is always compressed and encrypted regardless of these settings.

## Summary by Sourcery

Honor configured cloud storage encryption and compression options when creating backup metadata.

Bug Fixes:
- Fix cloud storage encryption and compression settings so explicitly disabled options are honored instead of always enabling both protections.

Enhancements:
- Allow cloud storage metadata encryption and compression flags to be set directly.

Tests:
- Add coverage verifying enabled, disabled, and absent cloud storage protection options produce the expected metadata flags.


ch_backup/logic/table.py
```diff
@@ -70,10 +70,10 @@ def backup(
 
         backup_name = context.backup_meta.get_sanitized_name()
 
-        if context.cloud_conf.get("cloud_storage", {}).get("encryption", True):
+        if context.cloud_conf.get("encryption", True):
```
aalexfvk: Nevertheless, it looks like logic is broken because `cloud_storage` contains `_encryption` and `_compression` as `True` by default.

ch_backup/logic/table.py
```diff
@@ -70,12 +70,12 @@ def backup(
 
         backup_name = context.backup_meta.get_sanitized_name()
 
-        if context.cloud_conf.get("cloud_storage", {}).get("encryption", True):
-            logging.debug('Cloud Storage "shadow" backup will be encrypted')
-            context.backup_meta.cloud_storage.encrypt()
-        if context.cloud_conf.get("cloud_storage", {}).get("compression", True):
-            logging.debug('Cloud Storage "shadow" backup will be compressed')
-            context.backup_meta.cloud_storage.compress()
+        context.backup_meta.cloud_storage.encrypted = context.cloud_conf.get(
```
aalexfvk: Log messages were missed

aalexfvk (approved):

PR #371, opened 2026-09-21
MDB-43707: Full cloud storage backups

roaldm153:
# Full backup of object storage data

## Problem

A backup of an object storage disk stores only the local metadata - files that
point at blobs in the source bucket. When those blobs are deleted, the backup is
worthless.

## Solution

A new option, `--copy-cloud-storage-data` for `ch-backup backup`
(`cloud_storage.copy_data` in the config), makes ClickHouse copy the data itself
into the backup bucket. Such a backup is restored without
`--cloud-storage-source-bucket` and survives the loss of the source data.

The option is off by default, so nothing changes for existing setups.

## What it is made of

1. **The option itself** - a temporary disk pointing at the backup bucket, the
   copy on backup, the restore from it.
2. **Copying in parallel** - the copy runs in the freeze thread pool instead of
   table by table.
3. **Deduplication of the copies** - parts already stored in an earlier backup
   are not copied again, their metadata keeps a link to that backup. Without
   this a daily backup of an unchanged table copies the whole disk every day,
   because `FREEZE` hardlinks every active part on each run.

## Worth knowing

- The data is copied by ClickHouse as is: unlike the metadata, it is neither
  compressed nor encrypted by ch-backup.
- Object storage data of a backup is kept while another backup links to it, so
  the space is freed when the last link is gone, not when the backup is deleted.
- ClickHouse forms the object storage directory name itself and replaces `-` with
  `_`, so backups whose names differ only in those characters would share one path.
  Such a name is now rejected when the backup is created.
- Cleanup of temporary disks used to return right after removing the
  configuration of the first one, so the rest were left in the ClickHouse
  configuration directory, and none of them were unregistered. All of them are
  removed now, followed by a single `SYSTEM RELOAD CONFIG`.

## Testing

- `uv run pytest tests/unit` - 452 tests.
- 12 integration scenarios in `cloud_storage_full_backup.feature`. They run as
  a part of the usual integration test suite and can be run alone with
  `make test-integration BEHAVE_ARGS="--tags=object_storage_copy"`.
- Not covered by the tests: objects larger than the threshold of a single
  `CopyObject`, which ClickHouse has to copy in several parts. Such an object
  does not fit into a test environment, so this is worth a look on a real
  installation before the option is enabled there.

## Size of the change

`git diff 17a0d81...HEAD`, 28 files, +3725 / −289.

| Area | Added | % of added lines |
|---|---:|---:|
| Production code (`ch_backup/`) | 1087 | 29% |
| Unit tests | 1981 | 53% |
| Integration scenarios (`.feature`) | 550 | 15% |
| Integration step definitions and helpers | 88 | 2% |
| ClickHouse configuration used by the tests | 19 | <1% |

Production code by file:

| File | Added | Removed |
|---|---:|---:|
| `clickhouse/disks.py` | 587 | 151 |
| `logic/table.py` | 182 | 31 |
| `backup/layout.py` | 118 | 18 |
| `ch_backup.py` | 43 | 2 |
| `backup/deduplication.py` | 42 | 1 |
| `clickhouse/control.py` | 40 | 1 |
| `backup/metadata/cloud_storage_metadata.py` | 25 | 1 |
| `clickhouse/models.py` | 16 | — |
| `cli.py` | 14 | — |
| `backup/metadata/backup_metadata.py` | 10 | 2 |
| `backup/metadata/__init__.py` | 5 | 1 |
| `config.py` | 5 | — |


ch_backup/ch_backup.py
```diff
@@ -570,13 +582,23 @@ def _delete(
                 return backup.name, None
 
             logging.info("Removing non-shared backup data parts")
+            cloud_data_shared = backup.cloud_storage.enabled and (
+                self._cloud_storage_data_is_shared(backup, dedup_references)
+            )
```
kirillgarbar: This variable can be created later inside `if backup.cloud_storage.enabled:`

ch_backup/clickhouse/disks.py
kirillgarbar: `clickhouse disks` might ignore the exception and just write it to the stderr, returning 0 as exit code: https://github.com/ClickHouse/ClickHouse/blob/2c8ef6fb24185fd3ecc8bb108cd34b6c0190d3b9/programs/disks/DisksApp.cpp#L256. Let's check that response is starting with `Error: ` and throw exception in case copying of cloud storage data fails.

ch_backup/logic/table.py
kirillgarbar: Full cloud storage backups may be incompatible with `use_inplace_cloud_restore` feature. This should be tested.

ch_backup/clickhouse/disks.py
```diff
@@ -314,68 +425,393 @@ def _os_copy(
...
+    """
+    credentials = storage_config["credentials"]
+    disk_config.pop("metadata_path", None)
+    disk_config["endpoint"] = _backup_storage_endpoint(storage_config, key_prefix)
+    disk_config["access_key_id"] = credentials["access_key_id"]
+    disk_config["secret_access_key"] = credentials["secret_access_key"]
+
+    proxy = _backup_storage_proxy(storage_config)
+    if proxy:
+        disk_config["proxy"] = proxy
```
kirillgarbar: If there is no proxy in ch-backup config but proxy exists in CH config then result config may be incorrect. It is also possible that there may be other config parameters incompatible to backup storage. I think creating new config by copying fixed set of parameters from original CH config will be safer.

ch_backup/logic/table.py
```diff
@@ -209,10 +244,13 @@ def _backup(
                                         context,
                                         freezed_table,
                                         backup_name,
+                                        backup_disks,
                                     )
                                     self._backup_cloud_storage_metadata(
-                                        context, freezed_table
+                                        context, copy_pool, freezed_table, backup_disks
                                     )
```
kirillgarbar: This method copies data instead of metadata and should be renamed

ch_backup/logic/table.py
```diff
@@ -96,16 +102,39 @@ def backup(
...
+            backup_disks = None
+            if context.backup_meta.cloud_storage.data_copied:
+                copy_workers = multiprocessing_config.get(
+                    "cloud_storage_backup_workers", 1
+                )
+                if copy_workers > 1 and not context.ch_ctl.ch_version_ge("23.3"):
+                    logging.warning(
+                        "It is unsafe to use cloud_storage_backup_workers > 1 with clickhouse version < 23.3"
+                        f" (cloud_storage_backup_workers: {copy_workers}, ch_version: {context.ch_ctl.get_version()})"
+                    )
```
kirillgarbar: It is better to throw if it is unsafe

ch_backup/logic/table.py
```diff
@@ -96,16 +102,39 @@ def backup(
...
-                context,
-                db,
-                db_tables[db.name],
-                backup_name,
-                schema_only,
-                multiprocessing_config,
-                change_times,
-            )
+        with ExitStack() as stack:
+            backup_disks = None
```
kirillgarbar: If stack is used for only one object using nullcontext or something else will be simpler

ch_backup/clickhouse/disks.py
```diff
@@ -37,12 +48,124 @@ class ClickHouseDisksException(RuntimeError):
...
+                )
+
+        _render_disks_config(
+            _get_config_path(self._config_dir, tmp_disk_name),
+            disks_config,
+        )
+        self._ch_ctl.reload_config()
+
+        disk = self._ch_ctl.get_disk(tmp_disk_name)
+        self._created_disks[tmp_disk_name] = disk
```
kirillgarbar: Could we create the temporary backup disks before starting the worker pools? create_disk() calls reload_config() and get_disk() through the shared ClickhouseClient, whose requests.Session is not thread-safe. Creating the disks upfront keeps ClickHouse configuration and client access in the main thread, while workers only run clickhouse-disks copy.

ch_backup/logic/table.py
kirillgarbar: Shadow directories of local disk are removed after backup. For cloud storage backups it was necessary to keep the links from shadow files to keep backups valid. For full cloud storage backup it is not the case, so shadow directories should be also removed.

ch_backup/clickhouse/control.py
```diff
@@ -1222,8 +1223,11 @@ def scan_frozen_parts(
...
             part_path = dir_entry.path
-            checksum = _get_part_checksum(part_path)
             rel_paths = list_dir_files(part_path)
+            if disk.keeps_object_metadata:
+                checksum = _get_cloud_part_checksum(part_path, rel_paths)
+            else:
+                checksum = _get_part_checksum(part_path)
             abs_paths = [Path(part_path) / file for file in rel_paths]
 
             size = calc_aligned_files_size(abs_paths, alignment=BLOCKSIZE)
```
kirillgarbar: Size is calculated based on local file size. It will be later added to total backup size. For cloud storage parts without data copying it is expected behaviour, but I am not sure if is ok for full backups. It should probably be discussed with the team

kirillgarbar (changes requested):

kirillgarbar:
@sourcery-ai review

